PR #1362, opened 2014-12-29
MODE-2401 Updated the query & indexing SPI to handle the "noquery" type attribute.

hchiorean:


modeshape-jcr/src/main/java/org/modeshape/jcr/spi/index/IndexChangeSetListeners.java
```diff
@@ -1,456 +0,0 @@
...
-        private void clearPropertyBuffer() {
-            for (int i = 0; i != numberOfProperties; ++i) {
-                newProps[i] = null;
-                oldProps[i] = null;
-            }
-            hasNewProps = false;
-            hasOldProps = false;
-        }
-    }
-}
```
hchiorean: nope, `IndexChangeSetListeners` doesn't appear to be used anywhere.

rhauch:
Looks good.

PR #1364, opened 2015-01-06
MODE-2402 Changed the WF kit to use a transient binary store when no store is explicitly configured.

hchiorean:


rhauch:
Looks great.

PR #1366, opened 2015-01-07
MODE-2403 Added a test case to validate that the LIKE operator works correctly and is case-sensitive.

hchiorean:


rhauch:
Looks good.

PR #1367, opened 2015-01-07
MODE-2367:Error to access contents of older versions via CMIS

okulikov:
This PR contains:
- Fix for the error accessing previous versions of the document;
- Test case for this scenario

### DIFF
--- a/web/modeshape-web-cmis/src/main/java/org/modeshape/cmis/JcrMsVersionNode.java
+++ b/web/modeshape-web-cmis/src/main/java/org/modeshape/cmis/JcrMsVersionNode.java
@@ -1,0 +1,102 @@
+/*
+ * ModeShape (http://www.modeshape.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


web/modeshape-web-cmis/src/main/java/org/modeshape/cmis/JcrMsRepository.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.chemistry.opencmis.jcr.JcrRepository;
+import org.apache.chemistry.opencmis.jcr.JcrTypeManager;
+import org.apache.chemistry.opencmis.jcr.PathManager;
+import org.apache.chemistry.opencmis.jcr.type.JcrTypeHandlerManager;
+
+/**
+ *
+ * @author kulikov
+ */
+public class JcrMsRepository extends JcrRepository {
```
hchiorean: what's the purpose of this class ? it doesn't seem to be doing anything different than the base class.

okulikov:
@hchiorean , removed that class. I used to debug the problem.

hchiorean:
looks good.
@rhauch  ?

web/modeshape-web-cmis/src/main/java/org/modeshape/cmis/JcrMsVersionNode.java
```diff
@@ -0,0 +1,102 @@
...
+        try {
+            return version.getIdentifier();
+        } catch (RepositoryException e) {
+            return null;
+        }
+    }
+
+    @Override
+    protected boolean isLatestVersion() throws RepositoryException {
+        return true;
```
rhauch: Is this correct? Always the latest version?

okulikov: Yes, it is correct. This class belongs to the not versioned nt:version node  type.

rhauch:
Looks good.

PR #1369, opened 2015-01-09
MODE-2406 Updated Wildfly to 8.2 (together with some transitive dependencies) and jboss-parent to 16.

hchiorean:


rhauch:
Looks good.

PR #1371, opened 2015-01-13
MODE-2408 Fixed the handling of empty ACLs

hchiorean:
If a node has empty ACLs, then the nearest non-empty ones in the hierarchy should be taken into account instead of assuming all permissions are granted on that node.

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/AccessControlManagerImpl.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/AccessControlManagerImpl.java
@@ -131,7 +131,7 @@ public class AccessControlManagerImpl implements AccessControlManager {
         }
 
         if (!hasPrivileges(absPath, new Privilege[] {privileges.forName(Privilege.JCR_READ_ACCESS_CONTROL)})) {
-            throw new AccessDeniedException();
+            throw new AccessDeniedException(JcrI18n.permissionDenied.text(absPath, "read access control content"));
         }
 
         AccessControlList acl = findAccessList(absPath, false);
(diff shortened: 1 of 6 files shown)


modeshape-jcr/src/main/java/org/modeshape/jcr/AccessControlManagerImpl.java
```diff
@@ -261,9 +262,9 @@ private JcrAccessControlList findAccessList( String absPath,
         SessionCache sessionCache = session.cache();
         Map<String, Set<String>> permissions = startingNode.getPermissions(sessionCache);
         CachedNode node = startingNode;
-        if (permissions == null && searchParents) {
-            // walk up the hierarchy until we get a set of permissions or we reach the root or a missing parent
-            while (permissions == null) {
+        // walk up the hierarchy until we get a set of non-empty permissions or we reach the root or a missing parent
+        if ((permissions == null || permissions.isEmpty()) && searchParents) {
```
rhauch: Minor nit: couldn't this line just be `if ( searchParents ) {` since the `permission`-related criteria would be handled in the `while` loop?

modeshape-jcr/src/main/java/org/modeshape/jcr/security/acl/AccessControlEntryImpl.java
```diff
@@ -158,6 +156,16 @@ public boolean equals( Object other ) {
 
     @Override
     public int hashCode() {
-        return this.principal.hashCode();
+        int result = principal.hashCode();
+        result = 31 * result + privileges.hashCode();
+        return result;
```
rhauch: Or maybe just the following?

    return HashCode.compute(principal,privileges);

rhauch:
A few minor nits noted above, but otherwise looks great.

hchiorean:
amended the commit based on the previous comments.

rhauch:
Looks great.

PR #1372, opened 2015-01-16
MODE-2409 Changed the ClusteredChangeBus to wait until it receives loopback-changesets.

hchiorean:
This is the only way (apart from not sending loopback messages through JGroups) to ensure that when a thread calls `bus.notify` all the "in-thread" listeners have actually been notified before continuing with that thread. This in turn is a precondition that must hold true for any changes in the system area to work correctly.


rhauch:
This is a very clever way to handle this thorny problem. Changes look good.

PR #1373, opened 2015-01-19
MODE-2410 Deprecated the "writeAdditionalDocument" method in the connector SPI.

hchiorean:


rhauch:
This is a bit more than deprecation, but it's still okay considering the implementation probably didn't work "correctly" anyway.

hchiorean:
I agree that it's more than deprecation, but IMO we don't want anyone using this method (while at the same time not breaking the API within a minor version)

rhauch:
Right. Looks good.

PR #1374, opened 2015-01-21
MODE-2412 Fixed the handling of large extracted text values for the database binary stores

hchiorean:
If the length of the text is larger than the size of the DB column, it will be trimmed.

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/Database.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/Database.java
@@ -81,6 +81,7 @@ public class Database {
     protected static final String STATEMENTS_FILENAME_SUFFIX = "_database.properties";
     protected static final String DEFAULT_STATEMENTS_FILE_PATH = STATEMENTS_FILE_PATH + STATEMENTS_FILE_PREFIX + "default"
                                                                  + STATEMENTS_FILENAME_SUFFIX;
+    protected static final int DEFAULT_MAX_EXTRACTED_TEXT_LENGTH = 1000;
 
     private static final Logger LOGGER = Logger.getLogger(Database.class);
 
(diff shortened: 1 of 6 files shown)


modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/Database.java
```diff
@@ -155,11 +159,29 @@ protected Database( Connection connection,
...
 
         initializeStatements();
         initializeStorage(connection);
     }
 
+    private int determineMaxExtractedTextLength( DatabaseMetaData metaData ) {
+        ResultSet resultSet = null;
+        try {
+            resultSet = metaData.getColumns(null, null, null,  EXTRACTED_TEXT_COLUMN_NAME);
+            return resultSet.next() ? resultSet.getInt("COLUMN_SIZE") : DEFAULT_MAX_EXTRACTED_TEXT_LENGTH;
```
rhauch: This will find all of the columns with that name in any table in any schema in any database, right? And won't it be possible to return more than one result? After seeing this code, this seems pretty brittle.

hchiorean: the only additional filtering criteria that's available is "tableName". However, using & passing that as a simple string (param) yields no result (i.e. no columns found) on certain DBs. For example: H2.

I'm happy to take any pointers as to how that param should be used with `dbMetadata`.

hchiorean:
added additional filtering based on table name.

modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/Database.java
```diff
@@ -158,6 +162,30 @@ protected Database( Connection connection,
...
+        ResultSet resultSet = null;
+        try {
+            String tableName = this.tableName;
+            if (metaData.storesLowerCaseIdentifiers()) {
+                tableName = tableName.toLowerCase();
+            } else if (metaData.storesUpperCaseIdentifiers()) {
+                tableName = tableName.toUpperCase();
+            }
+            resultSet = metaData.getColumns(null, null, tableName,  EXTRACTED_TEXT_COLUMN_NAME);
+            return resultSet.next() ? resultSet.getInt("COLUMN_SIZE") : DEFAULT_MAX_EXTRACTED_TEXT_LENGTH;
```
rhauch: Great.

rhauch:
Looks great.

PR #1375, opened 2015-01-21
MODE-2360: Add content hash property

okulikov:
This PR introduces cmis:contentStreamHash property

### DIFF
--- a/web/modeshape-web-cmis/src/main/java/org/modeshape/cmis/JcrMsVersion.java
+++ b/web/modeshape-web-cmis/src/main/java/org/modeshape/cmis/JcrMsVersion.java
@@ -16,26 +16,35 @@
 package org.modeshape.cmis;
 
 import java.util.Iterator;
+import java.util.Set;
 import javax.jcr.Node;
+import javax.jcr.Property;
 import javax.jcr.RepositoryException;
 import javax.jcr.UnsupportedRepositoryOperationException;
 import javax.jcr.version.Version;
 import javax.jcr.version.VersionException;
 import javax.jcr.version.VersionHistory;
 import javax.jcr.version.VersionIterator;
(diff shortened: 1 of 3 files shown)


rhauch:
Looks good to me, as long as that hex hash string conforms to the right CMIS format.

okulikov:
@rhauch , yes hash string follows CMIS format: {hash-alg}hash-string

web/modeshape-web-cmis/pom.xml
```diff
@@ -80,7 +80,6 @@
         <dependency>
             <groupId>org.modeshape</groupId>
             <artifactId>modeshape-jcr</artifactId>
-            <scope>test</scope>
```
hchiorean: @okulikov: the code should be using `org.modeshape.jcr.api.Binary` and then this change would not be needed - i.e. it should still be possible to compile against `modeshape-jcr-api` only

web/modeshape-web-cmis/src/main/java/org/modeshape/cmis/JcrMsVersion.java
```diff
@@ -84,4 +93,28 @@ public JcrVersion getVersion(String name) {
...
+    /**
+     * Get the hexadecimal form of the SHA-1 hash of the contents. 
+     * 
+     * @return the hexadecimal form of hash, or a null string if the hash could not be computed or is not known
+     */
+    private String getHash() {
+        try {
+            Node contentNode = getContextNode();
+            Property data = contentNode.getProperty(Property.JCR_DATA);
+            AbstractBinary bin = (AbstractBinary) data.getBinary();
```
hchiorean: see my previous comment

okulikov:
@hchiorean , fixed

PR #1376, opened 2015-01-26
MODE-2411 Extended the default JAAS provider to be able to interact with the JBoss container's SecurityManagementService.

hchiorean:
This should make sure whatever caching is configured via the container is being leveraged by ModeShape.


rhauch:
Looks good.

PR #1377, opened 2015-01-26
MODE-2384: CMIS web application does not work with clustering config

okulikov:
This PR adds cmis-web to the list of deployable web applications


okulikov:
Git failure

PR #1378, opened 2015-01-27
MODE-2413 Fixed the Base64 encoder to be able to handle SelfClosingInputStream (and in general any wrapped stream which might be closed independently)

hchiorean:


rhauch:
Looks good.

PR #1380, opened 2015-01-28
MODE-2411 Changed the warn message to a debug message when an unknown Credentials implementation is being passed in.

hchiorean:


rhauch:
Looks good.

PR #1381, opened 2015-01-29
MODE-2416 Fixed smartGWT dependency group id.

hchiorean:


rhauch:
How did this even work? Anyway, it looks good.

hchiorean:
it worked because interestingly enough the previous artifact is available in the JBoss Nexus (both only there) via the Third-party uploads repository

rhauch:
Ah, makes sense. I only checked Maven Central.

PR #1384, opened 2015-02-03
MODE-2387 Added support for additional MapDB configuration options.

hchiorean:


modeshape-jcr/src/main/java/org/modeshape/jcr/index/local/LocalIndexProvider.java
```diff
@@ -138,7 +148,33 @@ protected void doInitialize() throws RepositoryException {
                            file.getAbsolutePath());
         }
         // Get the database ...
-        this.db = DBMaker.newFileDB(file).make();
+        DBMaker<?> dbMaker = DBMaker.newFileDB(file);
+        if (this.cacheSize != null) {
+            dbMaker.cacheSize(cacheSize);
+            logger().trace("MapDB cache size set to {0}", cacheSize);
```
rhauch: What do you think about including the provider name in these messages?

rhauch:
Looks great, especially if provider name added to debug messages (see above).

hchiorean:
amended the commit

rhauch:
Thanks. Looks great.

PR #1385, opened 2015-02-06
MODE-2418 Fixed potential ConcurrentModificationException

hchiorean:
This PR contains 2 commits:
- one which adds a small optimization & a better exception message for the `JcrVersionManager`
- one which fixes a potentially much more critical issue: when using local transactions (i.e. started & committed by ModeShape during session.save) there was a potential leakage of nodes that were being changed/edited in that transaction into the shared workspace cache (used by everyone else incl. readers), __before the transaction was committed__ The solution was to make sure that a `TransactionalWorkspaceCache` is used *always*, regardless whether a user or a local transaction in active.


rhauch:
Looks good.

PR #1386, opened 2015-02-09
MODE-2424 Fixed locking validation when attempting to add a mixin and optimized some of the lock checks to make sure that if it exists, a lock is not retrieved multiple times.

hchiorean:


rhauch:
Looks good.

PR #1388, opened 2015-02-11
MODE-2418 Fixed the order in which the inner sessions are saved during checkin, making sure that the system session is saved first making the system-area data "available" first.

hchiorean:


rhauch:
Looks good.

PR #1389, opened 2015-02-12
MODE-2425 Fixed PATH() queries with SET criteria.

hchiorean:


modeshape-jcr/src/test/java/org/modeshape/jcr/JcrQueryManagerTest.java
```diff
@@ -2468,6 +2468,24 @@ public void shouldBeAbleToCreateAndExecuteJcrSql2QueryWithSubqueryInCriteria2()
...
+        String sql = "SELECT [car:maker] FROM [car:Car] WHERE PATH()" +
+                     " IN ('/Cars/Hybrid/Toyota Prius', '/Cars/Hybrid/Toyota Highlander', '/Cars/Hybrid/Nissan Altima')";
+        Query query = session.getWorkspace().getQueryManager().createQuery(sql, Query.JCR_SQL2);
+        QueryResult result = query.execute();
+        final List<String> expectedMakers = new ArrayList<>(Arrays.asList("Toyota", "Toyota", "Nissan"));
+        validateQuery().rowCount(3).onEachRow(new Predicate() {
+            @Override
+            public void validate( int rowNumber, Row row ) throws RepositoryException {
+                String actualValue = row.getValue("car:maker").getString();
+                expectedMakers.remove(actualValue);
```
okulikov: 'Toyota' appears twice in the list. I guess here is a bit of risk to remove both entries when actualValue equals 'Toyota' and skip another selected by query value, say 'Chevrolet'

hchiorean: wdym ?

okulikov: you are right. just cheked that arraylist.remove() removes first object and not both.

modeshape-jcr/src/main/java/org/modeshape/jcr/query/engine/ScanningQueryEngine.java
```diff
@@ -1537,6 +1538,9 @@ protected RowFilter createFilter() {
                     return new DynamicOperandFilter(operation) {
                         @Override
                         protected boolean evaluate( Object leftHandValue ) {
+                            if (Path.class.isAssignableFrom(defaultType.getType())) {
+                                leftHandValue = pathFactory.create(leftHandValue);
```
rhauch: Does this work properly if `leftHandValue` is an array?

hchiorean: According to [the API](https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/main/java/org/modeshape/jcr/value/ValueFactory.java#L472) it should.

rhauch: Yes, but that's not the method this code is calling. Since `leftHandValue` is of type `Object`, then `ValueFactory.create(Object)` will be the method that gets called, and the implementation doesn't currently check for array.

hchiorean: good point, I'll add the check

hchiorean:
amended the commit

rhauch:
Do the tests still work?

hchiorean:
yes, why ?

rhauch:
You changed it quickly, and I didn't know if you re-ran the tests.

hchiorean:
I only re-ran the `JcrQueryManagerTest` test case, but I don't really expect other test to fail because of the change.

rhauch:
Ok, great. Looks great.

PR #1391, opened 2015-02-16
MODE-2430 Fixed BsonDataInput stream truncating strings.

hchiorean:


rhauch:
Looks good. Nice catch, BTW. I'm sure it was not easy to find.

PR #1392, opened 2015-02-20
Added test to demonstrate MODE-2435

lucasweb78:
JCR-SQL2 queries that contain both an order by clause and a limit (with offset) clause are incorrectly returning results when the offset exceeds the total result set size.

The following PR contains a test that demonstrates this first by using limit and offset without a order by and then repeating the same query with an order by. The first query works as expected but the second fails returning rows when it shouldn't.


hchiorean:
Superseded by https://github.com/ModeShape/modeshape/pull/1394.

PR #1393, opened 2015-02-23
MODE-2432 Updated the index manager (both from the API & the SPI) to be able to expose information about that status of each configured index.

hchiorean:


modeshape-jcr-api/src/main/java/org/modeshape/jcr/api/index/IndexManager.java
```diff
@@ -117,5 +118,49 @@ void registerIndexes( IndexDefinition[] indexDefinitions,
...
+     * @param indexName a {@link String} the name of an index which should exist for the provider; may not be null
+     * @param workspaceName  a {@link String} the name of a workspace for which the index was created; may not be null
+     * @return a {@link org.modeshape.jcr.api.index.IndexManager.IndexStatus} instance, never {@code null} but possibly
+     * {@link org.modeshape.jcr.api.index.IndexManager.IndexStatus#NON_EXISTENT} if a provider with this name does not exist
+     * or an index with this name does not exist for the provider or the index doesn't exist for the given workspace.
+     */
+    IndexStatus getIndexStatus( String providerName, String indexName, String workspaceName );
 
+    /**
+     * Returns a list with the names of all the indexes that have a certain status, for a provider & workspace.
```
rhauch: Replace `&` with either `&amp;` or `and`.

modeshape-jcr/src/main/java/org/modeshape/jcr/index/local/ManagedLocalIndex.java
```diff
@@ -85,4 +93,17 @@ public void removeAll() {
...
     }
 
+    @Override
+    public IndexManager.IndexStatus getStatus() {
+        return status.get();
+    }
+
+    @Override
+    public void updateStatus(IndexManager.IndexStatus currentStatus, IndexManager.IndexStatus newStatus) {
+        this.status.compareAndSet(currentStatus, newStatus);
```
rhauch: So what happens if the `compareAndSet` method returns false and does nothing?

hchiorean: nothing :smile: 
The reason I'm using `compareAndSet` and not `set` is that I expect to only be able to move between certain statuses in a predefined order. This method could return a `boolean`, but callers wouldn't be using it

rhauch: It would be fine to not check it if there were only 2 states, but with more states it's important to know if one of the unexpected state transitions happens. If we don't, then should that happen what state will it be left in?

What do you think about either making an assertion after the `compareAndSet` to ensure that it was set indeed set properly, or throw an exception if `compareAndSet` returned `false`?

hchiorean: I'm not sure especially because of the "highly" multi-threaded nature of the indexes:  for example the `enable` & `disable` methods are present on the `ManagedIndex` SPI, but atm. their only use is a limited case of reindexing. By limited case I mean in a very particular scenario (at startup for example) indexes will go through `enabled` -> `disabled` sequences while they're being reindexed.
On the other hand, there is no SPI support for what's happening when `workspace.reindex` is called, even though it's semantically the same thing as above.

Introducing assertions or exceptions in this method would essentially mean "only moving between state A and state B" is correct, anything else is wrong. We could do that, but based on the current existing SPI it's totally unclear to me what those correct transitions are. So the implementation of this method at is "happy case" attempt of certain transitions I know for a fact can occur:
* enabled
* enabled - reindexing
* reindexing - enabled
* enabled - disabled

rhauch: Well, my concern is that the state gets left in some unusable state (e.g., reindexing) when it is no longer at that state, simply because the compare part of the `compareAndSet` failed. 

I'm not overly concerned, and I'm fine if you want to leave it as is.

rhauch:
Once the ampersand is fixed (see above), this could be merged.

PR #1396, opened 2015-03-16
MODE-2409 Removed code added as a previous fix for this issue & changed the JGroups config to achieve the same effect.

hchiorean:


rhauch:
Looks good.

PR #1397, opened 2015-03-16
MODE-2235: Web-explorer should show metrics

okulikov:
This PR implements charts for the repository metrcis

### DIFF
--- a/web/modeshape-web-explorer/pom.xml
+++ b/web/modeshape-web-explorer/pom.xml
@@ -57,7 +57,10 @@
             <artifactId>gwt-user</artifactId>
             <scope>provided</scope>
         </dependency>
-
+        <dependency>
+            <groupId>com.googlecode.gwtmosaic</groupId>
+            <artifactId>gwt-incubator</artifactId>
+        </dependency>
 
         <dependency>
             <groupId>com.smartgwt</groupId>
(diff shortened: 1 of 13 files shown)


web/modeshape-web-explorer/pom.xml
```diff
@@ -57,7 +57,12 @@
             <artifactId>gwt-user</artifactId>
             <scope>provided</scope>
         </dependency>
-
+        <dependency>
+            <groupId>com.googlecode.gwtmosaic</groupId>
+            <artifactId>gwt-incubator</artifactId>
+            <version>0.4.0-rc2</version>
+            <type>jar</type>
+        </dependency>
```
hchiorean: This should be declared in the parent's `<dependencyManagement>` section. `jar` is not required, it's the default type.
Also, isn't there a final version of the dependency we can use ? `rc2` makes me very nervous, not only because it's not final but also because this dependency is not available in Maven Central - which makes it __very exotic__ IMO

okulikov: Makes sense but unfortunately I did not find anything final available. This dependency is used only for single class `Canvas` which allows to draw on the page.

hchiorean: @okulikov: ok, in this case just pls make sure the parent defines the dependency.

hchiorean:
looks good

hchiorean:
@rhauch thoughts ?

rhauch:
Looks good.

PR #1399, opened 2015-03-18
MODE-2444 Teiid DDL Importer Processor Error: "Unparsable table body constraint" when importing from SAP HANA

elvisisking:
There were 2 issues: (1) INDEX was used as a column name in a column definition but was being parsed as a table element constraint, and (2) the parser was not handling embedded quotes in string literals. The parser was modified to handle the way the SQL spec allows for escaping embedding quotes (which is 2 consecutive quotes). Also the parser was changed to look ahead at the next token when it finds a keyword like INDEX when parsing table bodies. Added test cases and the full HANA DDL. This DDL file is huge so only the statements that test the 2 issues are left uncommented.


hchiorean:
looks good, thanks @elvisisking

hchiorean:
@rhauch thoughts ?

rhauch:
Looks good.

PR #1401, opened 2015-03-23
MODE-2346 Fixed the way commits are handled by git tree: they should not use "git log" to resolve the commit but instead rely on the branch provided to the git tree.

hchiorean:


connectors/modeshape-connector-git/src/main/java/org/modeshape/connector/git/GitTree.java
```diff
@@ -85,26 +85,23 @@ public Document execute( Repository repository,
...
-                    writer.addProperty(JcrLexicon.CREATED_BY, committer);
-                    writer.addProperty(GitLexicon.OBJECT_ID, objId.name());
-                    writer.addProperty(GitLexicon.AUTHOR, author);
-                    writer.addProperty(GitLexicon.COMMITTER, committer);
-                    writer.addProperty(GitLexicon.COMMITTED, committed);
-                    writer.addProperty(GitLexicon.TITLE, commit.getShortMessage());
-                    writer.addProperty(GitLexicon.HISTORY, GitHistory.referenceToHistory(objId, branchOrTagOrObjectId, values));
-                    writer.addProperty(GitLexicon.DETAIL, GitCommitDetails.referenceToCommit(objId, values));
-                }
+                // Add the properties for this node ...
```
rhauch: Is it still possible for `commit` to be null?

hchiorean: from the JavaDoc of `parseCommit`
>Returns:
reference to the commit object. Never null.

Not sure why the `null` checks were ever there...

rhauch: IIRC, it was because sometimes the commit was indeed null, despite the JavaDocs saying otherwise.

rhauch:
This is fine. I'm a bit leery about taking out the null checks, but I haven't worked with this code for a long time and don't exactly remember the circumstances.

hchiorean:
amended the commit and added back the `null` checks, if there is a possibility that the Javadoc is wrong...

PR #1408, opened 2015-04-20
MODE-2443-b: Enhance backup/restore process for the web explorer.

okulikov:
This PR contains the following changes:
- backup/restore dialogs with process parameters;
- backup/restore process moved into servlet's temporary directory.

### DIFF
--- a/web/modeshape-web-explorer/pom.xml
+++ b/web/modeshape-web-explorer/pom.xml
@@ -28,6 +28,7 @@
         <dependency>
             <groupId>org.modeshape</groupId>
             <artifactId>modeshape-jcr-api</artifactId>
+            <scope>provided</scope>
         </dependency>
 
         <dependency>
--- a/web/modeshape-web-explorer/pom.xml
+++ b/web/modeshape-web-explorer/pom.xml
@@ -28,7 +28,6 @@
         <dependency>
             <groupId>org.modeshape</groupId>
             <artifactId>modeshape-jcr-api</artifactId>
-            <scope>provided</scope>
         </dependency>
 
         <dependency>
--- a/web/modeshape-web-explorer/src/main/java/org/modeshape/web/BackupExportServlet.java
+++ b/web/modeshape-web-explorer/src/main/java/org/modeshape/web/BackupExportServlet.java
@@ -25,6 +25,7 @@ import javax.servlet.ServletContext;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import org.modeshape.common.logging.Logger;
 import org.modeshape.common.util.FileUtil;
 import org.modeshape.common.util.IoUtil;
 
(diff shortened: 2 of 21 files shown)


web/modeshape-web-explorer/src/main/java/org/modeshape/web/BackupExportServlet.java
```diff
@@ -34,18 +35,19 @@
...
 
     private static final long serialVersionUID = 1L;
-
+    private File tempDir;
+    
     @Override
     protected void doGet(HttpServletRequest request, HttpServletResponse response)
             throws IOException {
-        File file = new File("zzz");
+        File file = new File(tempDir.getAbsoluteFile() + File.separator + "zzz");
```
rhauch: Probably would be good to add something (e.g., something time-related) to the temporary file to help eliminate chances to clash if 2+ separate users each instigated a backup at overlapping times.

rhauch:
My comment above applies to multiple places.

okulikov:
@rhauch , @hchiorean Latest commit creates unique directories based on time for backup/restore

web/modeshape-web-explorer/pom.xml
```diff
@@ -28,7 +28,6 @@
         <dependency>
             <groupId>org.modeshape</groupId>
             <artifactId>modeshape-jcr-api</artifactId>
-            <scope>provided</scope>
```
hchiorean: i'm curios - why the need for this change ? How is this dependency different than `javax.jcr` for example, which is still `provided` ?

web/modeshape-web-explorer/src/main/java/org/modeshape/web/server/impl/RestoreUsrOptions.java
```diff
@@ -0,0 +1,42 @@
...
+package org.modeshape.web.server.impl;
+
+import org.modeshape.jcr.api.RestoreOptions;
+import org.modeshape.web.shared.RestoreParams;
+
+/**
+ *
+ * @author kulikov
+ */
+public class RestoreUsrOptions extends RestoreOptions {
```
hchiorean: what's the purpose of this class ? It seems to do nothing more than extend `RestoreOptions`

hchiorean:
@okulikov: as mentioned in my [JIRA comment](https://issues.jboss.org/browse/MODE-2443?focusedCommentId=13059115&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13059115), I'm not 100% clear why there was the need to create the `BackupParams` and `RestoreParams` classes, which essentially mirror (but with state) the `BackupOptions` and `RestoreOptions` from the API.
Why isn't is possible to code against these abstract classes directly and then pass in anonymous inner objects as actual values ?

web/modeshape-web-explorer/pom.xml
okulikov: That's just IDE trick when it appends imports. Missed this change.

okulikov:
@hchiorean ```BackupParams``` and ```RestoreParams``` are value objects used to pass user input to server side. It is impossible to use existing abstract ```BackupOptions``` and ```RestoreOptions``` for user input on the client side. Those parameters are not shareable between server side and client side.  The next pair of classes ```BackupUsrOptions``` and ```RestoreUsrOptions``` provide a glue (on the server side) between respective _parameters_ and _options_. That's why I was commenting on JIRA about API usage.

hchiorean:
tx for explaining @okulikov, looks good

rhauch:
Looks good.

PR #1409, opened 2015-04-21
MODE-2454 Changed the WF kit to use Infinispan as local modules(libraries) and not as a subsystem.

hchiorean:


rhauch:
Looks great.

PR #1410, opened 2015-04-24
MODE-2462 log problems when they exist

weebl2000:
I think this should be logged when problems is not empty.


hchiorean:
looks good, thanks.

PR #1411, opened 2015-04-30
MODE-2423 Moved to ISPN 7.2.0.CR1

hchiorean:


deploy/jbossas/kit/jboss-wf8/domain/configuration/modeshape/sample-ha-cache-config.xml
```diff
@@ -1,76 +1,68 @@
...
+                    isolation="READ_COMMITTED"
+                    acquire-timeout="1000"/>
+            <transaction
+                    mode="NON_XA"
+                    locking="PESSIMISTIC"/>
+            <persistence>
+                <file-store
+                        shared="false"
+                        preload="false"
+                        purge="true"
```
rhauch: This is set to `true` whereas it was `false` in the 6.x configuration. Is that intentional?

hchiorean: yes, if neither `preload` nor `purge` are set to `true` ISPN issues an ugly warning in replicated mode.

rhauch:
Looks good. Mostly configuration changes (which is to be expected since that changed dramatically), and very limited code changes.

PR #1412, opened 2015-04-30
MODE-2411 Fixed the setting of the securityDomain attribute.

hchiorean:


rhauch:
Were there other places (e.g., config files) where `domainName` was used and should now be `securityDomain`?

hchiorean:
no; this is a member of the `EnvironmentAuthenticationProvider` class which is set via reflection at repository startup.

rhauch:
Okay. Looks good.

PR #1413, opened 2015-05-01
MODE-2464: Ensures all input stream in Vdb Sequencers are closed

phantomjinx:
* VdbDynamicSequencer
 * Closes the initial input stream taken from the inputProperty's
   binaryValue

* VdbManifest
 * Closes the XML StreamReader instance on completion of its reading

* Closing these ensures that the streams are closed and releases any locks
  created on the original binary property value, avoiding the possibility
  of future deadlocks when saving the property.

Is it possible to backport this to a release of 4.2.x??

### DIFF
--- a/sequencers/modeshape-sequencer-teiid/src/main/java/org/modeshape/sequencer/teiid/VdbManifest.java
+++ b/sequencers/modeshape-sequencer-teiid/src/main/java/org/modeshape/sequencer/teiid/VdbManifest.java
@@ -1091,19 +1091,26 @@ public class VdbManifest implements Comparable<VdbManifest> {
                                  final Context context ) throws Exception {
             VdbManifest manifest = null;
             final XMLInputFactory factory = XMLInputFactory.newInstance();
-            final XMLStreamReader streamReader = factory.createXMLStreamReader(stream);
+            XMLStreamReader streamReader = null;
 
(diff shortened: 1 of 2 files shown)


hchiorean:
looks good

sequencers/modeshape-sequencer-teiid/src/main/java/org/modeshape/sequencer/teiid/VdbManifest.java
```diff
@@ -1091,19 +1091,26 @@ public VdbManifest read( final InputStream stream,
...
             VdbManifest manifest = null;
             final XMLInputFactory factory = XMLInputFactory.newInstance();
-            final XMLStreamReader streamReader = factory.createXMLStreamReader(stream);
+            XMLStreamReader streamReader = null;
 
-            if (streamReader.hasNext()) {
-                if (streamReader.next() == XMLStreamConstants.START_ELEMENT) {
-                    final String elementName = streamReader.getLocalName();
+            try {
+                streamReader = factory.createXMLStreamReader(stream);
```
rhauch: Hmm, I guess `XMLStreamReader` is not `Closable`? That's unfortunate.

rhauch:
Looks good.

PR #1416, opened 2015-05-06
MODE-2423 Moved to ISPN 7.2.0.Final

hchiorean:


boms/modeshape-bom-embedded/pom.xml
```diff
@@ -366,7 +366,7 @@
             </dependency>
             <!--Required by infinispan-cachestore-jdbc -->
             <dependency>
-                <groupId>c3p0</groupId>
+                <groupId>com.mchange</groupId>
```
rhauch: I'm curious, did this c3p0 dependency change in Infinispan between CR1 and Final?

rhauch:
Looks good.

boms/modeshape-bom-embedded/pom.xml
hchiorean: no, in CR1 ISPN was using 0.9.5-pre-X which was a bit shabby. So I postponed changing our version until ISPN uses a "final" c3p0 version.

PR #1419, opened 2015-05-12
MODE-2465 Changed DdlTokenizer to ignore "$"

freakolowsky:
Oracle, MySQL and postgres dialects consider a dollar sign as
a valid schema object name character. 

Table name was already handled previously but all other cases 
failed as subsection of ddl statement is created out of a list 
of tokens (at least in Oracle dialect).

Example:
Column definition like "COL$D VARCHAR2" was tokenized as
"COL", "$", "D", "VARCHAR2" and when rebuilt from tokens it
became "COL $ D VARCHAR2".
All subsequent parsing failed.

This is a generic solution.

Test result:
Tests run: 456, Failures: 0, Errors: 0, Skipped: 1

### DIFF
--- a/sequencers/modeshape-sequencer-ddl/src/main/java/org/modeshape/sequencer/ddl/DdlTokenStream.java
+++ b/sequencers/modeshape-sequencer-ddl/src/main/java/org/modeshape/sequencer/ddl/DdlTokenStream.java
@@ -374,7 +374,6 @@ public class DdlTokenStream extends TokenStream {
                     case '+':
                     case '%':
                     case '?':
-                    case '$':
                     case '[':
                     case ']':
                     case '!':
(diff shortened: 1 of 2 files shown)


hchiorean:
looks good

rhauch:
Are you sure that removing the `$` character from the list of token separators won't change any of the other sequencers?

freakolowsky:
i've checked oracle, postgres and mysql docs ... "$" is a valid object name character.

i've alslo ran all of the supplied unit tests in the module and none have failed ... so i would say it should not break any sequencers.

hchiorean:
@rhauch I ran a full build and there are no regressions; are you ok with merging this ?

rhauch:
Yeah, I'm good. Thanks for double checking for me.

PR #1421, opened 2015-05-19
MODE-2401 Fixed the fact that queryable children which were under parents that had the noquery attribute were not taken into account by the query engine.

hchiorean:


rhauch:
Looks good.

PR #1422, opened 2015-05-19
Fix typo in modeshape-explorer

weebl2000:
Fixed a typo.


rhauch:
Thanks for the catch. +1

web/modeshape-web-explorer/src/main/java/org/modeshape/web/server/impl/TimeUnit.java
```diff
@@ -28,7 +28,7 @@
         new TimeUnit(Window.PREVIOUS_60_SECONDS, "Last minute"),
         new TimeUnit(Window.PREVIOUS_60_MINUTES, "Last hour"),
         new TimeUnit(Window.PREVIOUS_24_HOURS, "Last day"),
-        new TimeUnit(Window.PREVIOUS_7_DAYS, "Last weak"),
+        new TimeUnit(Window.PREVIOUS_7_DAYS, "Last week"),
         new TimeUnit(Window.PREVIOUS_52_WEEKS, "Last half year"),
```
rhauch: Actually, this is not right either. `PREVIOUS_52_WEEKS` should be "Last year" (not "half" year).

PR #1423, opened 2015-05-19
Fix typo in modeshape-explorer

weebl2000:


hchiorean:
@weebl2000: pls squash the commits; it doesn't make sense to have multiple commits considering the fix. thanks.

PR #1425, opened 2015-05-19
Fixed typo in web-explorer.

weebl2000:


web/modeshape-web-explorer/src/main/java/org/modeshape/web/server/impl/TimeUnit.java
```diff
@@ -28,8 +28,8 @@
         new TimeUnit(Window.PREVIOUS_60_SECONDS, "Last minute"),
         new TimeUnit(Window.PREVIOUS_60_MINUTES, "Last hour"),
         new TimeUnit(Window.PREVIOUS_24_HOURS, "Last day"),
-        new TimeUnit(Window.PREVIOUS_7_DAYS, "Last weak"),
-        new TimeUnit(Window.PREVIOUS_52_WEEKS, "Last half year"),
+        new TimeUnit(Window.PREVIOUS_7_DAYS, "Last week"),
+        new TimeUnit(Window.PREVIOUS_52_WEEKS, "Last year"),
```
rhauch: While we're trying to correct these, I'd like us to make them even more precise. "Last week" generally means the previous full week, whereas "Previous 7 days" (as in the enum literal) typically means something very different. 

For example, on Tuesday May 19, 2015 saying "last week" probably means something like the period from Sunday May 10 through Saturday May 16. But another possible interpretation is within a Monday-Friday work context, which would mean the period from Monday May 11 through Friday. It's even different for a Sunday-Thursday context.

However, on Tuesday May 19, 2015 saying "Previous 7 days" (generally) implies one thing: the 7 days that immediately precede the current day. (Ok, it's not entirely precise, since one could mean the preceding 168 hours.)

So can we change these to "Previous ..." or "Preceding ..." ? Specifically:

    private static final TimeUnit UNITS[] = {
        new TimeUnit(Window.PREVIOUS_60_SECONDS, "Previous 60 seconds"),
        new TimeUnit(Window.PREVIOUS_60_MINUTES, "Previous 60 minutes"),
        new TimeUnit(Window.PREVIOUS_24_HOURS, "Previous 24 hours"),
        new TimeUnit(Window.PREVIOUS_7_DAYS, "Previous 7 days"),
        new TimeUnit(Window.PREVIOUS_52_WEEKS, "Previous 52 weeks"),

Thoughts?

weebl2000:
I agree that is more clear. Will make a new PR later, omw home now.

PR #1426, opened 2015-05-19
Changed timeunits' names to better reflect their definition

weebl2000:


weebl2000:
I should be more awake before I make these...

rhauch:
Rather than close these to open another, it's better to simply use `git push -f ...` to the branch; the PR will be updated with the new commit (overwriting the old commit), but we don't have all the overhead of all these bad PRs. It's okay for this, but next time please keep a single PR until it is correct.

weebl2000:
Thanks! I am new to contributing on github.

PR #1428, opened 2015-05-21
Fixed typo

weebl2000:


hchiorean:
tx for the fix, but that text is flawed and we'll have to revisit it ;)

weebl2000:
I agree that a distributed infinispan cache is NOT appropriate for Modeshape usage. We have tried it, but it leads to horrible inconsistencies.

rhauch:
@hchiorean, go ahead and merge. We need to fix this file anyway like you did for the documentation.

weebl2000:
Btw, use case where infinispan might work in distributed mode is where there is no shared cache store. Have not tried though.

PR #1430, opened 2015-05-29
MODE-2401 Added support for handling mixins which have the "noquery" type attribute

hchiorean:
Also, updated the code to reflect the differences between the `noquery` types and the previously existing ModeShape `$queryable` document attribute which has a different semantic: the first are defined by the JCR spec while the second were introduced internally by ModeShape to avoid indexing entire branches in the system area. After that, their semantics was exposed to the connectors as well.

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/connector/filesystem/FileSystemConnector.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/connector/filesystem/FileSystemConnector.java
@@ -924,7 +924,7 @@ public class FileSystemConnector extends WritableConnector implements Pageable {
                                                                            connector.lastModifiedTimeFor(file));
             // there is no way to observe the previous value, so fire <null>
             connectorChangeSet.propertyChanged(contentId, JcrNtLexicon.RESOURCE, Collections.<Name>emptySet(), contentId, null,
-                                               modifiedProperty, queryable);
+                                               modifiedProperty);
 
             // fire PROPERTY_CHANGED for nt:file/jcr:content/jcr:data/jcr:lastModifiedBy
             String owner = connector.ownerFor(file);
@@ -932,7 +932,7 @@ public class FileSystemConnector extends WritableConnector implements Pageable {
                 Property modifiedByProperty = connector.propertyFactory().create(JcrLexicon.LAST_MODIFIED_BY, owner);
                 // there is no way to observe the previous value, so fire <null>
                 connectorChangeSet.propertyChanged(contentId, JcrNtLexicon.RESOURCE, Collections.<Name>emptySet(), contentId,
-                                                   null, modifiedByProperty, queryable);
+                                                   null, modifiedByProperty);
             }
 
             try {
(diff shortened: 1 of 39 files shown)


rhauch:
Looks good.

weebl2000:
Question: it seems from looking at the code that a node will not be indexed when either the primary type has noquery set, or one of the mixins has noquery set. Wouldn't it be more complete to not index the node as the type which has noquery set and perhaps exclude the attributes from the mixin. I understand this would be a lot more complex. Also I don't know what the JCR spec says about the noquery attribute.

hchiorean:
>Wouldn't it be more complete to not index the node as the type which has noquery set and perhaps >exclude the attributes from the mixin. I understand this would be a lot more complex.
I don't understand what you mean. A node is indexed as a whole; you also need to keep in mind that mixins are dynamic - you can add and remove them at runtime and therefore the node has to behave accordingly.

>Also I don't know what the JCR spec says about the noquery attribute
The only thing that the spec says is that if `noquery` is present on a node type, that node type should be queryable

weebl2000:
Nevermind, I think the current approach is the way to go. I was thinking the query/noquery attribute could be set differently on different mixins (or primary types), thus preventing a node which has two mixins set, one with and one without noquery to not be found using the query on the node which does not have noquery set.

The noquery attribute is primarily used for performance gain, I'm guessing, so letting noquery take precedence seems like the right approach.

weebl2000:
I'm going to add another comment, however this is probably not the right place for the discussion. Currently the attribute noquery implemented in Modeshape is more aptly described as not indexing the node type. This is the behaviour I would like to see and I am very glad with this commit. However, other users might want to have noquery behave like that, in a sense, just to not be able to query that kind of node, but still having it indexed as another node type. However, I don't see any specific use cases for this at the moment. Does anyone know of any benefits other than not indexing a node (and saving resources) when setting noquery?

Thanks.

hchiorean:
>However, other users might want to have noquery behave like that, in a sense, just to not be able to query that kind of node, but still having it indexed as another node type.
I'm not sure I follow: the semantics of `noquery` is pretty clearly defined in the JCR spec. ModeShape is a JCR  implementation. The fact that ModeShape 4 chooses to expose different types of indexes, is purely an internal implementation details.

weebl2000:
Let's say we have

[mix:a] mixin query

[mix:b] mixin noquery

If I create a node which has both mixins, I would not be able to find it querying [mix:a]. I do not mind this, but perhaps others might?

hchiorean:
You should not be using these mixins on the same node in the first place; it doesn't really make sense. 
One should design their node schema accordingly...

weebl2000:
Good point. :+1:

weebl2000:
@hchiorean @rhauch 

Could you point me into the right direction on where to document the behavior of noquery in Modeshape. Perhaps in other implementations a situation as I described above will produce results when querying [mix:a], so it is good to have this documented. 

The docs.jboss.org website pages seem a little broken by the way. I see Unknown macro: {space-metadata-from} everywhere.

hchiorean:
as mentioned previously, `noquery` _is not ModeShape specific_ but defined by the JCR spec: http://www.day.com/specs/jcr/2.0/25_Appendix.html#25.2%20Compact%20Node%20Type%20Definition%20Notation

> /* If 'query' is present then the node type is queryable. If 'noquery' is present then the node type is not queryable. If neither query nor noquery are present then the queryable setting of the node type is a variant. */
Query ::= ('noquery' | 'nq') | ('query' | 'q' )

variant = up to the implementation, not mandated by the spec. In ModeShape's case *everything* is queryable, unless `noquery` is defined.

weebl2000:
I know this, but it seems like a good idea to document what happens if both are assigned, no?

hchiorean:
we can document this, but assigning both at the same time is a clear indication that the client code "doesn't know what it's doing".

PR #1431, opened 2015-06-02
MODE-1460 Changed the code (as per ISPN-1987) to use a decorated cache which suppresses return values for the put & replace methods for which the ModeShape code doesn't care about the return value.

hchiorean:


modeshape-schematic/src/main/java/org/infinispan/schematic/internal/CacheSchematicDb.java
```diff
@@ -78,22 +81,15 @@ public boolean containsKey( String key ) {
...
-        return oldValue != null ? removedResult(key, oldValue) : null;
-    }
-
-    protected SchematicEntry removedResult( String key,
-                                            SchematicEntry entry ) {
-        SchematicEntryLiteral literal = (SchematicEntryLiteral)entry;
-        literal.markRemoved(true);
-        return literal;
+        // use storeForWriting because we don't care about the return type - i.e. we're doing a local put
+        storeForWriting.put(key, newEntry);
```
rhauch: This change doesn't maintain the old behavior of marking the old `SchematicEntryLiteral.removed = true`. Is that really not used anymore?

hchiorean: I couldn't find any usages for the `removed` flag (or the related methods). If this flag is somehow used/relevant I can revert those particular changes.

rhauch:
This looks okay pending a response to my question above. Also, is `put` used very much, and if so where? IIRC, the `LocalDocumentStore` (and `FederationStore`?) really use `SchemaEntry` editing rather than `put`, and that should be the majority of cases. I suspect that for all new entries we're using `putIfAbsent` so that adding the new entry is atomic.

modeshape-schematic/src/main/java/org/infinispan/schematic/internal/CacheSchematicDb.java
rhauch: I don't know that it was. I suspect it was leftover from the delta-aware stuff removed long ago.

hchiorean:
`put` is used (outside of the connectors) here: https://github.com/ModeShape/modeshape/blob/master/modeshape-schematic/src/main/java/org/infinispan/schematic/internal/CacheSchematicDb.java#L150
This in turn is used in several places.

New entries are indeed added via `putIfAbsent` which preserves the old behavior.

rhauch:
Hmmm, that reference to the `[editContent(...)](https://github.com/ModeShape/modeshape/blob/master/modeshape-schematic/src/main/java/org/infinispan/schematic/internal/CacheSchematicDb.java#L150)` seems like a potential race condition. Should that use `putIfAbsent` instead?

hchiorean:
I don't follow: the edit operation (either on the Local or on the Federated document store) are only used internally via the private `persistChanges`, after ISPN locks have been obtained on the relevant node keys. 
The `DocumentStore` interface is/was never meant to be used directly, so IMO there shouldn't be any concurrency concerns.

rhauch:
> I don't follow: the edit operation (either on the Local or on the Federated document store) are only used internally via the private persistChanges, after ISPN locks have been obtained on the relevant node keys. 

Yes, except that when an entry does not exist, ISPN will not obtain any locks for it. Therefore, if multiple threads (in whatever processes) attempt to create/edit an entry with this code, they might both be able to do so. The question is whether we can be 100% sure that this never happens. For example, two different threads should never try to create a *new node* using the same key (since keys are randomly generated). But isn't this code used to obtain an editor for a node that is expected to exist. For example, `DocumentTranslator` ultimately uses this method in `changeChildren(...)` to create/edit the blocks of children info. Not sure if it's also used elsewhere.

Either we analyze how this code is actually used and very 100% that multiple threads can never call this with the same key (seems difficult to guarantee in all future changes), or we change this method to use `putIfAbsent` so that it can react correctly in the presumably rare case that this code block does get called.

And if we do that, and that's the only place `put` was actually called, then might it make sense to completely remove the `put` methods from Schematic?

rhauch:
Let me rephrase my concern. First, I'm doubtful that the changes described in MODE-1460 (and implemented here) will really have much impact because nearly all of the times SchematicEntries are created and modified will be via entry editing or `putIfAbsent` (neither of which should be affected by these proposed changes).

Second, what are the cases when `put` is actually used? Seems to me that such calls are always potential race conditions in a distributed and concurrent system, and we should really consider using `putIfAbsent` instead. And if we find that we should always using `putIfAbsent` and never `put`, then perhaps we should remove the `put` methods from Schematic.

I hope that makes sense.

hchiorean:
>Second, what are the cases when put is actually used? 

The main `put` use cases come via  https://github.com/ModeShape/modeshape/blob/master/modeshape-schematic/src/main/java/org/infinispan/schematic/internal/CacheSchematicDb.java#L154  which in turn is: https://github.com/ModeShape/modeshape/blob/master/modeshape-schematic/src/main/java/org/infinispan/schematic/internal/SchematicEntryLiteral.java#L166

>Seems to me that such calls are always potential race conditions in a distributed and concurrent system, and we should really consider using putIfAbsent instead.

I disagree: `literal.edit` is likely to be called just as many times as `putIfAbsent` for new nodes. The discussion above is really in an `if` block which reads: `if literal == null && createIfMissing`. In this particular case (`createIfMissing`) we could use `putIfAbsent`, but again this only relates to a very limited number of use cases.

rhauch:
Okay. I still think that within the `createIfMissing` case (which I agree occurs rarely) we should use `putIfAbsent` instead of `put`. However, I'm fine with all of the other changes.

hchiorean:
amended the commit and made the discussed change, when creating a new entry into the cache.

rhauch:
Thanks. Looks great, @hchiorean.

PR #1432, opened 2015-06-04
Small Sequencer Fixes found during testing

phantomjinx:
Problems/omissions encountered during debugging of the sequencers in development of VdbBuilder/komodo.

### DIFF
--- a/sequencers/modeshape-sequencer-ddl/src/main/java/org/modeshape/sequencer/ddl/dialect/teiid/TeiidDdlLexicon.java
+++ b/sequencers/modeshape-sequencer-ddl/src/main/java/org/modeshape/sequencer/ddl/dialect/teiid/TeiidDdlLexicon.java
@@ -87,6 +87,103 @@ public class TeiidDdlLexicon extends StandardDdlLexicon implements TeiidDdlConst
     }
 
     /**
+     * JCR names related to the teiid materialized options.
+     */
+    public interface MaterializedOptions {
+
+        /**
+         * Whether table is materialized
+         */
+        String MATERIALIZED = "MATERIALIZED";
+
+        /**
+         * Materialized table
+         */
+        String MATERIALIZED_TABLE = "MATERIALIZED_TABLE";
+
+        /**
+         * Whether table is updateable
+         */
+        String UPDATABLE = "UPDATABLE";
+
+        /**
+         * Allow Teiid based management
+         */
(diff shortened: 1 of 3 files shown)


sequencers/modeshape-sequencer-ddl/src/main/java/org/modeshape/sequencer/ddl/dialect/teiid/TeiidDdlLexicon.java
```diff
@@ -87,6 +87,103 @@
...
+
+        /**
+         * Whether table is updateable
+         */
+        String UPDATABLE = "UPDATABLE";
+
+        /**
+         * Allow Teiid based management
+         */
+        String ALLOW_MATVIEW_MANAGEMENT = "teiid_rel:ALLOW_MATVIEW_MANAGEMENT";
```
rhauch: Is `teiid_rel` a JCR namespace prefix, or something else? If a ns prefix, then it should use a constant like the other constants in this file.

phantomjinx: It is actually part of the teiid property name -> [External Materialization](https://docs.jboss.org/author/display/teiid810final/External+Materialization?_sscc=t).

rhauch: Then because these string values follow the conventions of JCR names (but are not to be treated as JCR names), does this work correctly? For example, the code might try to parse these into a JCR name. If so, then the `:` should be escaped.

hchiorean:
@phantomjinx: in addition to @rhauch's comments, pls open a JIRA for this & attach the PR. Also, if possible, squash the 3 commits into 1. Thanks.

sequencers/modeshape-sequencer-ddl/src/main/java/org/modeshape/sequencer/ddl/dialect/teiid/TeiidDdlLexicon.java
phantomjinx: The ddl sequencer has been written so that it encodes the ':' with session.encode() hence options named as such are sequenced successfully.

phantomjinx:
JIRA created and commits squashed.

hchiorean:

rhauch:
Looks good.

PR #1433, opened 2015-06-05
MODE-2469 Added a "maxPoolSize" configuration attribute which allows the configuration of the sequencer's and text extractor's maximum number of threads. 

hchiorean:
Updated the AS kit to expose configuration attributes both for the names of the thread pools and the maximum size of the thread pool.


rhauch:
Looks good.

PR #1434, opened 2015-06-10
MODE-2345 Updated event-related metrics using the new 4.x ring buffer.

hchiorean:


modeshape-jcr/src/main/java/org/modeshape/jcr/bus/RepositoryChangeBus.java
```diff
@@ -76,7 +101,11 @@ public boolean register( ChangeSetListener observer ) {
         if (observer == null || shutdown.get()) return false;
         try {
             registrationLock.lock();
-            return ringBuffer.addConsumer(observer);
+            boolean result = ringBuffer.addConsumer(observer);
+            if (result && statistics != null) {
+                statistics.increment(ValueMetric.LISTENER_COUNT);               
+            }      
+            return result;
```
rhauch: It probably would be better to use a no-op `RepositoryStatistics` so that `statistics` is never null. It'd simplify the logic in these calls a bit, plus allow the `statistics` field to be `final` (allow for some additional JIT optimizations).

rhauch:
Looks good, though I do have one suggestion for a minor improvement (see line note above).

hchiorean:
I'm not convinced for a number of reasons:
1. the only place where `null` can come in is the tests - and there are a lot of tests which use a `RepositoryChageBus` where statistics are irrelevant
2. `RepositoryStatistics` implements a non-mutable interface - `RepositoryMonitor`  while all the `set` methods are internal to `RepositoryStatistics`
3. `RepositoryStatistics` also implements `ChangeSetListener` so a null implementation would have to implement a lot of methods coming from multiple places

Even though a `nullable` interface can be created, I'm not convinced it's worth the complexity since it would only be used in tests.

rhauch:
That's fine.

PR #1437, opened 2015-06-22
MODE-2473 Fixed behavior of nodeIterator#skip when entire batches coming from an index provider are skipped.

hchiorean:


rhauch:
Looks good.

PR #1439, opened 2015-06-26
MODE-2481 Added some more optimization to the ChildReferences implementations

hchiorean:
The optimizations are in 2 different areas:
* the data structures being used to hold the node information - ModeShape's `MultiMap` family is heavyweight compared to the standard JDK `Maps` and doesn't require it being used in this case
* passed SNS information down to the `ChildReferences` implementations, allowing further optimizations in the way SNS indexes are computed in certain cases as well as the `Context` instance being used.


rhauch:
Looks good.

PR #1440, opened 2015-06-30
MODE-2445 Failed to import SalesForce Metadata

elvisisking:
The issue about parsing escaped single quotes had already been fixed since this issue was written. However, parsing this DDL revealed an issue with foreign key references. I've changed DdlSequencer to perform a second pass in order to resolve the forward-looking references. Also fixed a few javadoc warnings.


rhauch:
Looks good

PR #1441, opened 2015-06-30
MODE-2477 Static VDBs doesn't get sequenced

elvisisking:
I removed the mandatory attribute from vdb:path property in vdb:marker as it is not required. When it is not present the validation error source is the model.


rhauch:
Looks good.

PR #1443, opened 2015-07-08
Fix MODE-2490

jasperstein:
ORDER BY col1, col2 used to cause a NullPointerException whenever col1 or col2 were nullable and indeed did contain null values. This is because the row's col1 and col2 were being serialized as a tuple, and the serialization (correctly) assumed that since a tuple was created, the sort key was non-null. However serialization of the tuple was done by serializing each component, and they +can+ be null.

Therefore, the component serializers should use a null-safe wrapper around the original serializer, see their various getSerializer() implementations. The wrapper itself is in lines 585-615. Since I guess we don't always know what serializers we have, and some serializers also leave not enough wiggle room to accomodate NULLs without violating all kinds of contracts, we have to have (unfortunately) an extra sentinel boolean to indicate a null value (if set to false, lines 594-595 and 605).

The Tuples class holds a WeakHashMap to store the wrappers (line 617) so the same wrapper will be retrieved when requested multiple times. Perhaps there's a better way/place to store them; I am currently too unfamiliar with ModeShape to know. (Maybe inside the Serializers class?) By using a WeakHashMap I ensure that the wrappers will be removed whenever the bufferManager gets garbage-collected.

Whenever a null-safe serializer is requested, it is first retrieved from the WeakHashMap if available. If not, it is first checked whether the original serializer from the bufferManager is null-safe. If so, no wrapping is done and the original serializer is used.

Yours,
Jasper Stein (XebiaLabs)

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/index/local/MapDB.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/index/local/MapDB.java
@@ -65,6 +65,13 @@ public class MapDB {
                                                      Comparator<?> comparator,
                                                      boolean pack );
 
+        /**
+         * Obtain a serializer for the given value type that can handle nulls.
+         *
+         * @param type the type; may not be null
+         * @return the serializer that can handle nulls
+         */
+        Serializer<?> nullSafeSerializerFor( Class<?> type );
     }
 
     public static Serializers serializers( ValueFactories factories ) {
@@ -78,9 +85,14 @@ public class MapDB {
 
     public static final class SerializerSupplier implements Serializers {
         private final Map<Class<?>, Serializer<?>> serializersByClass;
+        private final Map<Class<?>, Serializer<?>> nullSafeSerializersByClass;
         private final Map<Class<?>, BTreeKeySerializer<?>> bTreeKeySerializersByClass;
+        private final Serializer<?> DEFAULT_NULL_SAFE_SERIALIZER;
 
         protected SerializerSupplier( ValueFactories factories ) {
+            DEFAULT_NULL_SAFE_SERIALIZER = (isNullSafe(DEFAULT_SERIALIZER) ? DEFAULT_SERIALIZER
+                                                                           : new NullSafeSerializer<>(DEFAULT_SERIALIZER));
+
             // Create the serializers ...
             serializersByClass = new HashMap<Class<?>, Serializer<?>>();
             serializersByClass.put(String.class, Serializer.STRING);
@@ -96,11 +108,24 @@ public class MapDB {
             serializersByClass.put(NodeKey.class, NODE_KEY_SERIALIZER);
 
             bTreeKeySerializersByClass = new HashMap<Class<?>, BTreeKeySerializer<?>>();
+            nullSafeSerializersByClass = new HashMap<Class<?>, Serializer<?>>();
             for (Map.Entry<Class<?>, Serializer<?>> entry : serializersByClass.entrySet()) {
                 Serializer<?> serializer = entry.getValue();
                 @SuppressWarnings( {"rawtypes", "unchecked"} )
                 BTreeKeySerializer<?> bTreeSerializer = new DelegatingKeySerializer(serializer);
                 bTreeKeySerializersByClass.put(entry.getKey(), bTreeSerializer);
+
+                Serializer<?> nullsafe = (isNullSafe(serializer) ? serializer : new NullSafeSerializer<>(serializer));
+                nullSafeSerializersByClass.put(entry.getKey(), nullsafe);
+            }
+        }
+
+        private static <T> boolean isNullSafe(Serializer<T> serializer) {
+            try {
+                serializer.serialize(new DataOutput2(), null);
+                return true;
+            } catch(IOException | NullPointerException e) {
+                return false;
             }
         }
 
@@ -109,6 +134,12 @@ public class MapDB {
             Serializer<?> result = serializersByClass.get(type);
             if (result != null) return result;
             return DEFAULT_SERIALIZER;
+        }
+
+        public Serializer<?> nullSafeSerializerFor( Class<?> type ) {
+            Serializer<?> result = nullSafeSerializersByClass.get(type);
+            if (result != null) return result;
+            return DEFAULT_NULL_SAFE_SERIALIZER;
         }
 
         @Override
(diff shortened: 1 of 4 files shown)


hchiorean:
thanks for the PR @jasperstein 

The proposed change will work, but I think it's a bit of an overkill (e.g. storing the serializers into a `WeakHashMap`) because the `Tuples` instances really just "wrap" core serializers, without doing any serialization logic per-se.

The current code already caches serializers for the default data types [here](https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/main/java/org/modeshape/jcr/index/local/MapDB.java#L83). At the moment however, we're using the default MapDB serializers which don't support `null` values, hence the reported issue.

IMO it would be easier & simpler instead of using the default MapDB serializers in the above class, to simply add a `NullSafeSerializer` there (like in the current PR), wrap each of those serializers and store the wrapper (i.e. the `NullSafeSerializer`) instead of the default MapDB serializer.

jasperstein:
Sure, that would certainly work. Make it so :-) The reason I did not put them there by default is that on single-column sorting, NULLs seem to be handled well enough. The null-safe wrappers add this sentinel boolean, which means the serialised form now takes more space. I'm not well enough aware of general ModeShape use and the use of these serializers to decide whether the extra space would be an issue for some users. Since I estimate that Tuples, in general, are used way less often than the more basic types, using null-safe wrappers only for Tuples would cause lower space impact. That's why I put them inside the Tuples class. But if that's not a concern, it would indeed be simpler to move them into the Serializers class as the default choice. Shall I do that, or will you? (Or is it a concern after all?)

Yours,
Jasper Stein

hchiorean:
to me that doesn't seem like such a big concern (the extra space required by the boolean flag), considering the fact that it simplifies the fix & code. It's actually a "side-effect" of the JCR spec - you can't have 'null' values on single properties - that this issue only surfaces for multi-valued props.

@rhauch thoughts ?

rhauch:
First of all, thanks @jasperstein for tracking this down, figuring out a solution, and submitting a pull-request. I'd say this proposal is pretty close to ideal -- except for how the handlers are managed in memory.

Both of you have great points. I agree with @hchiorean that the better place for these null-handling serializers is in the [MapDB class](https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/main/java/org/modeshape/jcr/index/local/MapDB.java) that already manages the other serializers. However, I agree with @jasperstein that we should not use the null-handling serializers everywhere because the extra boolean is (i) extra space and (ii) not backwards compatible with existing indexes. So how about a compromise? :-)

What if we add a new method to the `Serializers` class similar to the existing [serializerFor(Class<?>)](https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/main/java/org/modeshape/jcr/index/local/MapDB.java#L54) method, except named `serializerForNullable(...)` and which returns the null-handling serializer for the supplied class. This lets the places that use the existing serializers able to get the same serializers they've always used, but then @jasperstein's [changes to the various tuple factories](https://github.com/ModeShape/modeshape/pull/1443/files#diff-ec7ca035f4196046d302111de7b8642aL641) (which I think are perfect) can use these new null-handling serializers. The big thing is that the `SerializerSupplier` implementation class remains immutable, and we can avoid altogether the complexity of managing the null-handling serializers with a WeakHashMap, which was pretty problematic (see below). I think this approach gives us the best of both worlds: we limit the use of these null-handling serializers to only serializing tuples and do not change how singular values are serialized or handled, and we clean up the way the null-handling serializers are managed/stored.

If @hchiorean agrees, then can you, @jasperstein, implement these suggestions and update the PR? Please amend your existing commit on the branch, then push forcibly to your fork (e.g., `git push -f origin MODE-2490` with the appropriate remote) to update the PR.

On a side note, the use of WeakHashMap here would have been a problem if we were to use it. First, it is not thread safe, so multiple threads trying to find a null-handling serializer would have been problematic, and using `serialized` is too heavyweight. Second, that approach used the `BufferManager` as keys in that map, and `BufferManager` was not designed nor is safe to be used as keys in any kind of hash map.

hchiorean:
@rhauch sounds good

jasperstein:
Yes, this idea had already occurred to me also... Great minds think alike, I guess ;-) So do I really amend the current commit or shall I just push an extra commit onto this branch?

rhauch:
We certainly allow PRs to contain more than one commit, but that's usually when each commit advances the codebase to a useful state. In these cases, multiple commits is perfectly fine. The thing we try to avoid is having commits that advance the codebase to a state that we prefer not to have, either because that state represents some feature or fix that is incomplete.

Since your first commit added some things we rather not have, it's simply easier if you amend the commit on your PR. Remember, this kind of "rewriting history" is fine as long as nobody else uses that history, and its okay on your repository because you did the work on a branch. We'll end up merging only the new commit, and you (and everyone else) will see that on the 'master' branch.

Make sense?

jasperstein:
Certainly makes sense :-) But since rewriting history is usually warned against, I thought I'd double-check. I'm working on fixing it, but I can continue wednesday at the earliest. So I'm not gone, just a few (holi)days without access to my (work) laptop.

jasperstein:
OK, pushed forcibly :-) For some reason locally I have a failed GitConnectorTest, but it seems to be unrelated to my code. It also fails on my previous PR version, and even on master. Please confirm I haven't actually broken things.

To get the solution working, I had to extend the Serializers interface into a new one (never change a published interface :-) especially if you don't know, like me, where it's all being used). The extended interface is called NullSafeSerializers which may be a bit misleading but I couldn't come up with a better-and-still-concise name. I'm open for suggestions.

MapDB.serializers() now returns this interface and SerializerSupplier implements it; so I can actually query for null-safe serializers from within BufferManager (see new line numbers 335 and 456). The wrapper itself is now a private inner class of SerializerSupplier. Not sure if it makes sense to make it more public...? I hijacked the BTree serializers map filling loop to also fill the new NullSafe serializers map. See MapDB.java:120-122

Finally apparently some 'code cleanup' was performed to remove spaces at the end of some of the comment lines. My apologies, I just noticed.

rhauch:
Looks good, though I'm not convinced that introducing a new `NullSafeSerializers` interface is necessary. I can't think why anybody (outside of our codebase) would implement our `Serializers` interface for production code. This interface is primarily part of an API that we ourselves implement, so seems like we could add the new method to it.

hchiorean:
+1, I also think a new `NullSafeSerializers` interface is not necessary.

jasperstein:
OK, I folded the new interface into the old one, and all tests are passing (except, still, two of the GitConnectorTest ones that seem unrelated). Amend + force push again, I suppose?

hchiorean:
@jasperstein +1

jasperstein:
Here it is :-)

hchiorean:
looks good

rhauch:
Looks good. Nicely done, @jasperstein

jasperstein:
@hchiorean @rhauch Thanks for your compliments!

PR #1444, opened 2015-07-09
MODE-2484, MODE-2485 Fixed several potential ISPN locking issues

hchiorean:
This is the forward porting of the issues already fixed for MODE-2484 & MODE-2485.


rhauch:
Looks good.

PR #1445, opened 2015-07-10
MODE-2493 Fixed the rollback handling of ModeShape transactions.

hchiorean:
This is a forward-port of the existing fix.


rhauch:
Looks good

PR #1447, opened 2015-07-15
MODE-2494 Adding check whether a child node join has been reversed du…

dankelleher:
…ring query optimization, mirroring logic for SameNodeJoinCondition and EquiJoinCondition.

For some complex queries, such as:

SELECT parent.[jcr:path], child1.[jcr:name], desc.[jcr:name]
FROM [nt:unstructured] AS parent
LEFT OUTER JOIN [nt:unstructured] AS child1 ON ISCHILDNODE(child1,parent)
INNER JOIN [nt:unstructured] AS desc on ISCHILDNODE(desc, child1)
LEFT OUTER JOIN [nt:unstructured] AS child2 ON ISCHILDNODE(child2,parent)
WHERE ISCHILDNODE(parent,'/')
AND NAME(child2) = 'Hybrid'
AND NAME(desc) LIKE 'Nissan%'

the query optimizer reverses the left and right sides of the join, leading to an assertion error in ScanningQueryEngine.

This fix checks if the join has been reversed, and creates rowextractors for the opposite columns if this is the case.

Added three tests, two which are fixed by this change, and one which is still an issue, suggesting a deeper problem. For the remaining (Ignored) test, the exception is in the DescendantNodeJoinCondition code.
This seems less simple to resolve, and probably a true fix lies in ReplaceViews or somewhere similar.

### DIFF
--- a/modeshape-jcr/src/test/java/org/modeshape/jcr/JcrQueryManagerTest.java
+++ b/modeshape-jcr/src/test/java/org/modeshape/jcr/JcrQueryManagerTest.java
@@ -2642,6 +2642,56 @@ public class JcrQueryManagerTest extends MultiUseAbstractTest {
         validateQuery().rowCount(17).hasColumns("left.jcr:path").validate(query, result);
     }
 
+    @FixFor( "MODE-2494" )
+    @Test
+    public void shouldBeAbleToCreateAndExecuteJcrSql2QueryWithTwoLeftOuterJoinsOnIsChildNodeWithSubsequentIsChildNode() throws RepositoryException {
+        String sql = "SELECT parent.[jcr:path], child1.[jcr:name], desc.[jcr:name] FROM [nt:unstructured] AS parent " +
+                "LEFT OUTER JOIN [nt:unstructured] AS child1 ON ISCHILDNODE(child1,parent) " +
+                "INNER JOIN [nt:unstructured] AS desc on ISCHILDNODE(desc, child1) " +
+                "LEFT OUTER JOIN [nt:unstructured] AS child2 ON ISCHILDNODE(child2,parent) " +
+                "WHERE ISCHILDNODE(parent,'/') " +
+                "AND NAME(child2) = 'Hybrid' " +
+                "AND NAME(desc) LIKE 'Nissan%'";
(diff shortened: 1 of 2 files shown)


modeshape-jcr/src/test/java/org/modeshape/jcr/JcrQueryManagerTest.java
```diff
@@ -2642,6 +2642,56 @@ public void shouldBeAbleToCreateAndExecuteJcrSql2QueryWithLeftOuterJoinOnIsChild
...
+                "AND NAME(child2) = 'Hybrid' " +
+                "AND NAME(desc) LIKE 'Nissan%'";
+        Query query = session.getWorkspace().getQueryManager().createQuery(sql, Query.JCR_SQL2);
+        QueryResult result = query.execute();
+        validateQuery().rowCount(1).validate(query, result);
+    }
+
+    @FixFor( "MODE-2494" )
+    @Test
+    @Ignore("This is not fixed by the fix for MODE-2494, and points to a potentially deeper problem, " +
```
hchiorean: does this mean there's an additional issue even with this fix ? If yes, should that be a new JIRA ? If no, then there's little point in having an ignored test case

hchiorean:
with the exception of the above question, the fix looks good to me, thanks @dankelleher

modeshape-jcr/src/test/java/org/modeshape/jcr/JcrQueryManagerTest.java
dankelleher: Yes, I think it probably should be a separate JIRA. I plan to look into it more before creating the ticket... I still don't fully understand what the query optimizers are doing. The point behind the ignored test case was to have a placeholder test, for this new ticket (always a good place to start when fixing a bug!).

I saw a few other ignored tests in the suite, so decided to follow suit here, but if it is considered to be clutter, I'm happy to remove it.

hchiorean: we can keep it as is then; just wanted to make sure we're not missing another bug.

rhauch:
Looks good to me.

PR #1448, opened 2015-07-20
MODE-2488 Moved the AS kit to Wildfly 9

hchiorean:
There are 2 parts to the PR:
* one which updates some pieces of the ModeShape AS subsystem API, making sure that the current version of the subsystem __is binary compatible__ with Wildfly 8 while also running on Wildfly 9. This means that even though we're using the latest Wildfly API, we're using it in a backwards-compatible manner with Wildfly 8. This also means that some parts of the subsystem are using `deprecated` methods in Wildfly 9 because there is no equivalent in Wildfly 8. [This](https://github.com/wildfly/wildfly-core/blob/1.0.0.Final/controller/src/main/java/org/jboss/as/controller/AbstractAddStepHandler.java#L328) is the most notable example.

* one which updates all of the assemblies, configurations & artifacts so that those from Wildfly 9 are used.


rhauch:
Looks great.

PR #1449, opened 2015-07-27
MODE-2489 Added the ability to configure mime-type detection

hchiorean:
The current (magic content) based mime-type detection is still the default, but now users can configure the mime-type detection mechanism via:
```
repository {
  binaryStorage: {
    "mimeTypeDetection" : ["content", "name", "none"]   
  }
}
```
where `name` ignores the stream altogether (simply uses the name of the owning node) and `none` disables mime-type detection.

The AS kit (based on [Wildfly 9](https://github.com/ModeShape/modeshape/pull/1448)) was also updated to expose this functionality in the binary stores:
```
<file-binary-storage mime-type-detection="content|name|none"/>
```

This PR requires https://github.com/ModeShape/modeshape/pull/1448 to be merged first, since it affects the AS kit.


rhauch:
Looks good.

PR #1450, opened 2015-07-31
MODE-2495 Fixes how user transactions are handled

hchiorean:
The previous code made the incorrect assumption that a user transaction is "confined" to the thread that started it. This is obviously not the case, since the same transaction can be suspended & resumed subsequently off different threads.

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/txn/NoClientTransactions.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/txn/NoClientTransactions.java
@@ -50,14 +50,15 @@ public final class NoClientTransactions extends Transactions {
 
     @Override
     public synchronized Transaction begin() throws NotSupportedException, SystemException {
-        if (ACTIVE_TRANSACTION.get() == null) {
+        NestableThreadLocalTransaction tx = ACTIVE_TRANSACTION.get();
+        if (tx == null) {
             // Start a transaction ...
             txnMgr.begin();
             if (logger.isTraceEnabled()) {
                 logger.trace("Begin transaction {0}", currentTransactionId());
             }
-            ACTIVE_TRANSACTION.set(new NestableThreadLocalTransaction(txnMgr, ACTIVE_TRANSACTION));
+            tx = new NestableThreadLocalTransaction(txnMgr, ACTIVE_TRANSACTION);
         }
-        return ACTIVE_TRANSACTION.get().begin();
+        return tx.begin();
     }
 }
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/txn/SynchronizedTransactions.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/txn/SynchronizedTransactions.java
@@ -17,6 +17,7 @@ package org.modeshape.jcr.txn;
 
 import java.util.HashSet;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicReference;
 import javax.transaction.NotSupportedException;
 import javax.transaction.RollbackException;
@@ -39,12 +40,13 @@ import org.modeshape.jcr.cache.document.WorkspaceCache;
  */
 public final class SynchronizedTransactions extends Transactions {
 
-    private static final ThreadLocal<Transaction> ACTIVE_TRANSACTION = new ThreadLocal<Transaction>();
+    private static final ThreadLocal<NestableThreadLocalTransaction> LOCAL_TRANSACTION = new ThreadLocal<NestableThreadLocalTransaction>();
 
     @SuppressWarnings( "rawtypes")
     private final Cache localCache;
     private final AtomicReference<TransactionListener> transactionListener = new AtomicReference<>();
     private final TransactionTable transactionTable;
+    private final ConcurrentHashMap<Long, SynchronizedTransaction> activeTransactionsByISPNGlobalTxId = new ConcurrentHashMap<>();
 
     /**
      * Creates a new instance which wraps a transaction manager and monitor factory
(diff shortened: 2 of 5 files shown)


modeshape-jcr/src/main/java/org/modeshape/jcr/txn/NoClientTransactions.java
```diff
@@ -50,14 +50,15 @@ public Transaction currentTransaction() {
...
             // Start a transaction ...
             txnMgr.begin();
             if (logger.isTraceEnabled()) {
                 logger.trace("Begin transaction {0}", currentTransactionId());
             }
-            ACTIVE_TRANSACTION.set(new NestableThreadLocalTransaction(txnMgr, ACTIVE_TRANSACTION));
+            tx = new NestableThreadLocalTransaction(txnMgr, ACTIVE_TRANSACTION);
         }
-        return ACTIVE_TRANSACTION.get().begin();
+        return tx.begin();
```
rhauch: I'm not sure I understand these changes ... the thread-local `ACTIVE_TRANSACTION` is checked above, but it is no longer ever set within this class. (And this class is final, so there can be no subclasses that do set it.)

Either `ACTIVE_TRANSACTION` should be removed and this code cleaned up, or it should be fixed so that something sets it with a `NestableThreadLocalTransaction` reference.

modeshape-jcr/src/main/java/org/modeshape/jcr/txn/SynchronizedTransactions.java
```diff
@@ -66,35 +68,49 @@ public SynchronizedTransactions( TransactionManager txnMgr, Cache localCache ) {
...
-            // we have an existing transaction so depending on the type we either need to be aware of nesting
-            // or return as-is
+            // we have an existing local transaction so we need to be aware of nesting
             if (logger.isTraceEnabled()) {
                 logger.trace("Found active ModeShape transaction '{0}' ", result);
             }
-            return result instanceof NestableThreadLocalTransaction ?
-                   ((NestableThreadLocalTransaction) result).begin() :
-                   result;
+            return ((NestableThreadLocalTransaction) result).begin();
```
rhauch: This casting is no longer necessary, since `LOCAL_TRANSACTION` holds `NestableThreadLocalTransaction` references.

rhauch: BTW, I actually like how `LOCAL_TRANSACTION` now holds only the specific type, since that prevents holding any other kind of `Transaction` (like the `SynchronizedTransaction` created below when the transaction was not created in the current thread). And I realize that the `result` variable is of the more general type, but it'd be great if this logic could be a little cleaner. The only way I can think to do that is to do something like this:

    NestedThreadLocalTransaction ntlt = LOCAL_TRANSACTION.get();
    if ( ntlt != null ) {
        ...
        return ntlt.begin();
    }
    Transaction result = ntlt;

modeshape-jcr/src/main/java/org/modeshape/jcr/txn/Transactions.java
```diff
@@ -384,11 +384,12 @@ public void commit()
...
         private AtomicInteger nestedLevel = new AtomicInteger(0);
-        private ThreadLocal<? extends Transaction> txReference;
+        private ThreadLocal<NestableThreadLocalTransaction> txHolder;
 
-        protected NestableThreadLocalTransaction( TransactionManager txnMgr, ThreadLocal<? extends Transaction> txReference) {
+        protected NestableThreadLocalTransaction( TransactionManager txnMgr, ThreadLocal<NestableThreadLocalTransaction> txHolder ) {
             super(txnMgr);
-            this.txReference = txReference;
+            this.txHolder = txHolder;
+            this.txHolder.set(this);
```
hchiorean: this is were the transaction is set on the `ThreadLocal` reference

modeshape-jcr/src/main/java/org/modeshape/jcr/txn/NoClientTransactions.java
hchiorean: The active transaction is set in the constructor of `NestableThreadLocalTransaction` (see below).
The reason for the change is that both `NoClientTransactions` and `SynchronizedTransactions` have the exact same logic of setting the active transaction.

modeshape-jcr/src/main/java/org/modeshape/jcr/txn/SynchronizedTransactions.java
hchiorean: The casting is required because the method returns a `Transaction` instance, which is also the type of the local variable. On the other hand, `Transaction` does not have a `begin` method, since only local transactions can be "started".

modeshape-jcr/src/main/java/org/modeshape/jcr/txn/Transactions.java
```diff
@@ -384,11 +384,12 @@ public void commit()
 
     protected class NestableThreadLocalTransaction extends TraceableSimpleTransaction {
         private AtomicInteger nestedLevel = new AtomicInteger(0);
-        private ThreadLocal<? extends Transaction> txReference;
+        private ThreadLocal<NestableThreadLocalTransaction> txHolder;
```
rhauch: Hmm.... If a `NestableThreadLocalTransaction` is created in a certain thread, then by using `ThreadLocal` here the code below will only find a non-null reference when accessed from the same thread. If the transaction is resumed in a different thread, upon cleanup (when the code tries to clean up this `NestableThreadLocalTransaction`) the contents of the `txHolder` will not be visible to the current thread. I've not got my head wrapped around this, but it seems like we're not properly removing the reference in `txHolder` so that when a new transaction is started on the same thread that started our first transaction, it will find an existing `NestableThreadLocalTransaction` reference.

Having `NestableThreadLocalTransaction` hold onto the thread local seemed like a good idea when it was assumed that a local transaction would only be active on a single thread. But now that we know that's not a valid assumption, this class should no longer hold onto the thread local.

Something just seems not quite right.

modeshape-jcr/src/main/java/org/modeshape/jcr/txn/NoClientTransactions.java
rhauch: Right, but see my new line comment about potential cleanup problems with that code.

modeshape-jcr/src/main/java/org/modeshape/jcr/txn/SynchronizedTransactions.java
rhauch: Yes, which I why I suggested an alternative above that doesn't require the cast.

hchiorean: >NestedThreadLocalTransaction ntlt = LOCAL_TRANSACTION.get();
if ( ntlt != null ) {
    ...
    return ntlt.begin();
}
Transaction result = ntlt;

I understand what you're saying, but [this](https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/main/java/org/modeshape/jcr/txn/SynchronizedTransactions.java#L133) part of the code needs to always be executed because it adds extra (non trivial) logic for debugging purposes. Because of this, the method can't "short-circuit" and exit and has to make sure that part of the code is always executed, regardless of the transaction type.

modeshape-jcr/src/main/java/org/modeshape/jcr/txn/NoClientTransactions.java
hchiorean: The comment about the potential cleanup was for `SynchronizedTransactions`, this is `NoClientTransactions`. Not sure how that could apply here

modeshape-jcr/src/main/java/org/modeshape/jcr/txn/SynchronizedTransactions.java
rhauch: I don't follow. The result of `LOCAL_TRANSACTION.get()` can *only* be a `NestedThreadLocalTransaction`. Because of this, IIUC my suggestion does not change any logic whatsoever, and the tertiary operator on lines 82-83 reduces to a simple return.

modeshape-jcr/src/main/java/org/modeshape/jcr/txn/Transactions.java
hchiorean: `NestableThreadLocalTransaction`s are only used when there is no external transaction - i.e. when ModeShape starts & commits a transaction during `session.save()`. If that is the case, these type of transactions are "internal" to Modeshape and not visible on the outside, hence they cannot be context-switched.

Context switching on the other hand is specific/possible only to user-transactions (i.e. `SynchronizedTransaction` instances) which are not held anymore in `ThreadLocal` variables.

modeshape-jcr/src/main/java/org/modeshape/jcr/txn/SynchronizedTransactions.java
hchiorean: my bad, the previous comment (about debugging) applies only to newly started transactions, so the suggested change (which is really just to avoid the cast) is possible.

modeshape-jcr/src/main/java/org/modeshape/jcr/txn/NoClientTransactions.java
rhauch: The cleanup code I mentioned was in `NestedThreadLocalTransaction` [here](https://github.com/ModeShape/modeshape/pull/1450/files#diff-971a87033e57696cf8bfee58d9a4ce74R417), so that definitely applies.

modeshape-jcr/src/main/java/org/modeshape/jcr/txn/SynchronizedTransactions.java
rhauch: Well, it's not just about avoiding the cast. Rather, it's about making it very clear that the reference can only be of type `NestedThreadLocalTransaction` at this point. If you'd prefer, this is another way to do it:

    Transaction result = LOCAL_TRANSACTION.get();
    if ( result != null ) {
        ...
        return ((NestedThreadLocalTransaction)ntlt).begin();
    }

I guess this is really about getting rid of the unnecessary tertiary statement (and the type check); we can still keep the cast (though I think my previous suggestion is more static type friendly, should somebody change the type of the ThreadLocal). But either will work.

modeshape-jcr/src/main/java/org/modeshape/jcr/txn/Transactions.java
rhauch: *sigh*

modeshape-jcr/src/main/java/org/modeshape/jcr/txn/NoClientTransactions.java
hchiorean: >The cleanup code I mentioned was in NestedThreadLocalTransaction here, so that definitely applies.
not sure I understand what the cleanup issue is then
In the following code:
```
       NestableThreadLocalTransaction tx = ACTIVE_TRANSACTION.get();
1.        if (tx == null) {
2.             // Start a transaction ...
3.            tx = new NestableThreadLocalTransaction(txnMgr, ACTIVE_TRANSACTION);
4.         }
5.        return tx.begin();
```
on line 3. the `NestableThreadLocalTransaction` instance is set on the `ACTIVE_TRANSACTION` instance (via the constructor). 
However, once a transaction is committed/rolledback, we need to clear the reference which is the reason for [this code](https://github.com/ModeShape/modeshape/pull/1450/files#diff-971a87033e57696cf8bfee58d9a4ce74R417).

rhauch: I'm crossing line note streams. I'm fine with your rebuttal, and where the `ACTIVE_TRANSACTION` thread local reference is set.

modeshape-jcr/src/main/java/org/modeshape/jcr/txn/SynchronizedTransactions.java
rhauch: I think this is my only suggestion for changes ... to get rid of the unnecessary type check and tertiary statement (and keeping the cast). Granted, it's minor change, but IMO it adds some important clarity now that `LOCAL_TRANSACTION` is now typed to only hold `NestedThreadLocalTransaction`.

hchiorean:
Amended the commit with a broader refactoring of the `SynchronizedTransactions#begin` method.

rhauch:
Very nice improvement: the updated method is indeed more straightforward. Thanks! Looks great!

PR #1452, opened 2015-08-04
MODE-2496 Makes sure custom authentication providers in the AS kit are added first in the list

hchiorean:
This essentially makes sure that any custom providers will take precedence over the default (built-in) security providers which interface with the AS's security domain.


rhauch:
Looks good.

PR #1454, opened 2015-08-04
MODE-2459: Improve web-explorer ergonomic

okulikov:
This PR improves web explorer ergonomic and refactors source code to make it more strightforward.


hchiorean:
@okulikov: did a simple smoke test by building the PR and starting the application in the Wildfly 9 kit from the `integration/modeshape-jbossas-integration-tests/target/wildfly-9.0.0.Final` folder. Navigating to the `artifacts` repository and attempting a simple navigation (i.e. clicking on the `files` or `folders` nodes produces an error dialog - see attached screenshot).

So this is IMO a basic regression test which fails compared to the previous version.

artifact_repo

okulikov:
@hchiorean , I understood. I need to fix property editors for the multiple values properties

okulikov:
@hchiorean , commited fix.

hchiorean:
@okulikov: next simple test case - same repository
1. click on the edit icon of the `cars` node and enter `carrs`. Hit `Ok`
cars_edit1
2. "unidentified" dialog pops up, no stack trace on the server:
cars_edit2

I'm fully aware that validating the functionality of the Explorer would require a full QE effort - which we don't have- but we should at least make sure that some very basic test cases work out-of-the-box or we remove functionality from the UI. It makes no sense to me to have a UI full of non-working options.

Can you please make sure you do some simple smoke tests (manually by clicking around) on a repository which has data in it. Thanks.

okulikov:
@hchiorean , I need also add more accurate exception handing also. The common way to catch all exceptions and just show a message produces "empty" dialogs

okulikov:
@hchiorean , the new commit fixes discovered errors and others that I was able to catch as well. The only not good thing remains is that import/export works for server side file system. I guess would be more useful to use local file system as well as for backup/restore option. Do we want to modify this feature within current PR?

hchiorean:
>Do we want to modify this feature within current PR?

no, you can open a separate JIRA for that if you want.

hchiorean:
@okulikov: this latest version of the PR is still not functioning properly IMO:
- create a new `nt:folder` into the `artifacts` repository
- attempt to `save`

Clicking the `save` icon doesn't do anything and there is no way a user can actually persist his changes.

okulikov:
@hchiorean , fixed save. Will do another round of testing

okulikov:
@hchiorean , updated PR with fixes.

hchiorean:
@okulikov: same problem as before (cannot save) exists when `editing` a node (i.e. changing its name).

okulikov:
@hchiorean , fixed.

hchiorean:
@okulikov: this looks good now, with just 2 additional items:

1. https://issues.jboss.org/browse/MODE-2499 - can be handled separately from this PR
2. what is the purpose of the search bar (see screenshot) ?
search_explorer

- is it so search repositories ? If yes, then something isn't working correctly - i.e. if you enter the letter `a` all repositories (from `modeshape-jbossas-integration-tests`) are returned even though some don't have the letter `a` in their name.
- is it something else ?

okulikov:
@hchiorean , it searches using name and "descriptor" which is a combination of the vendor and version (This text appears under repository name). More specific query will narrow search results.

hchiorean:
@okulikov: ok, then I'm fine with merging this. Thanks.
@rhauch thoughts on this PR ?

rhauch:
Sounds good.

okulikov:
@hchiorean , I did squash commits.

hchiorean:
@okulikov: looks good, thanks

hchiorean:
merged in {{master}}.

PR #1456, opened 2015-08-07
MODE-2483 Allows custom authorization providers to be called for workspace based permissions

hchiorean:
The previous code would circumvent custom providers altogether, meaning that right after `session.login` any custom provider had to have assigned ModeShape-specific roles to the authenticated user in order for the login to be successful.


rhauch:
Looks good.

PR #1458, opened 2015-08-18
MODE-2467: Modeshape-explorer: add mixin when clicking refresh.

okulikov:
This PR is just a simple fix for a dialog window.


hchiorean:
looks good

PR #1459, opened 2015-08-25
MODE-2109 Adds support for unordered large collections

hchiorean:
These are a set of 4 new built-in node types and 1 abstract base type, which store their children into buckets based on the SHA1 of the name of each child. The essential algorithm and different collections types are described [in this comment](https://issues.jboss.org/browse/MODE-2109?focusedCommentId=12928688&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-12928688).

Any JCR operations where the path or name of a child node is known are a lot faster than for regular nodes (e.g. `session.getNode(path)` or `node.getNode(path)`). However, operations where only the ID of node is known - like `node.getPath()` *can be slower* since it requires looking for a node with a certain ID in each of the buckets incrementally. 

These new node types support large flat hierarchies, but they do have some limitations:
* SNS, ordering, renaming, versioning or federating external sources are not supported
* copying and cloning are only partially supported - i.e. you can copy/clone something into a large collection, but you cannot perform the operation if the source is a large collection

Other regular JCR operations like querying, import/export etc are supported in similar fashion to regular nodes.


modeshape-jcr/src/main/java/org/modeshape/jcr/NodeTypes.java
```diff
@@ -238,6 +240,10 @@ protected NodeTypes( ExecutionContext context,
                     nodeTypeNamesThatAreShareable.add(name);
                 }
 
+                if (nodeType.isNodeType(ModeShapeLexicon.UNORDERED_COLLECTION)) {
+                    nodeTypeNamesThatAreUnorderableCollections.add(name);
```
rhauch: How are the tiny, small, large, and huge unordered collection node types added to the `nodeTypeNamesThatAreUnorderableCollections `?

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/BucketId.java
```diff
@@ -0,0 +1,66 @@
...
+ * @see BucketedChildReferences
+ */
+class BucketId implements Comparable<BucketId> {
+    private final String id;
+
+    protected BucketId( Name name, int prb ) {
+        this(name.getString(), prb);
+    }
+
+    protected BucketId( String nameString, int prb ) {
```
rhauch: Obviously `prb` used to determine the length of the SHA1 that is used, but I can't figure out what `prb` stands for.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/BucketedChildReferences.java
```diff
@@ -0,0 +1,375 @@
...
+import org.modeshape.jcr.value.ValueFactory;
+
+/**
+ * A child references implementation which uses separate buckets for groups of children, based on the SHA1 of the name of each 
+ * child and the an order of magnitude constant (named 'prb') stored in the parent document. 
+ * See <a href='https://issues.jboss.org/browse/MODE-2109'>MODE-2109</a>
+ * 
+ * @author Horia Chiorean (hchiorea@redhat.com)
+ */
+class BucketedChildReferences extends AbstractChildReferences {
```
rhauch: Can this be annotated with `@Immutable`?

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/BucketId.java
```diff
@@ -0,0 +1,66 @@
...
+import org.modeshape.common.util.SecureHash;
+import org.modeshape.jcr.value.Name;
+
+/**
+ * Abstraction for the ID of a bucket.
+ * 
+ * @author Horia Chiorean (hchiorea@redhat.com)
+ * @see BucketedChildReferences
+ */
+class BucketId implements Comparable<BucketId> {
```
rhauch: Can this be annotated with `@Immutable`? Maybe even `final`?

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/BucketedChildReferences.java
```diff
@@ -0,0 +1,375 @@
...
+        ChildReference ref = bucket != null ? bucket.childReferenceFor(name) : null;
+
+        if (ref == null) {
+            // we don't have a reference in our data, look at the changes
+            Iterator<ChildInsertions> iterator = changes.insertions(name);
+            if (iterator.hasNext()) {
+                ChildInsertions inserted = iterator.next();
+                Iterator<ChildReference> iter = inserted.inserted().iterator();
+                // we don't support SNS, renames or reorderings so we'll always return the fist insertion
+                return iter.hasNext() ? iter.next() : null;
```
rhauch: How do we know that the first inserted child has the correct name?

modeshape-jcr/src/main/java/org/modeshape/jcr/NodeTypes.java
hchiorean: by virtue of the node type definitions: https://github.com/hchiorean/modeshape/blob/MODE-2109/modeshape-jcr/src/main/resources/org/modeshape/jcr/modeshape_builtins.cnd#L152
Any of the default and custom node types which are unordered collections are parsed when the node types are loaded.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/BucketId.java
hchiorean: `prb` is actually the term you [came up with](https://issues.jboss.org/browse/MODE-2109?focusedCommentId=12928688&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-12928688) :smile: 
it's essentially the size of the bucket. But you're right, I can rename this field to something more readable.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/BucketedChildReferences.java
hchiorean: +1, it should be annotated

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/BucketId.java
hchiorean: +1, it should be annotated and made final

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/BucketedChildReferences.java
hchiorean: `changes.insertions(name)` will only return insertions with that certain name. If those exist, they must be it because SNS are not supported.

modeshape-jcr/src/main/java/org/modeshape/jcr/NodeTypes.java
rhauch: Oh, sorry. I misread `nodeType.isNodeType` as comparing node type *names*. My bad. I can only claim early-morning syndrome.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/BucketId.java
rhauch: How about just `length` or something like that? At least in this class.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/BucketedChildReferences.java
rhauch: Great.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/DocumentConstants.java
```diff
@@ -46,6 +46,9 @@
     public static final String REFERENCE_FIELD = "$ref";
     public static final String WEAK_REFERENCE_FIELD = "$wref";
     public static final String SIMPLE_REFERENCE_FIELD = "$sref";
+    public static final String PRB = "$prb";
```
rhauch: BTW, it's okay to use something short as the literal. If you want to keep using `PRB`, then perhaps add a comment.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/DocumentTranslator.java
```diff
@@ -868,15 +868,22 @@ protected long insertChildren( EditableDocument document,
...
         boolean hasChildren = document.containsField(CHILDREN);
         boolean hasFederatedSegments = document.containsField(FEDERATED_SEGMENTS);
-        if (!hasChildren && !hasFederatedSegments) {
+        if (!hasChildren && !hasFederatedSegments && !isUnorderedCollection) {
             return ImmutableChildReferences.EMPTY_CHILD_REFERENCES;
         }
+
+        if (isUnorderedCollection) {
+            // the base 'mode:unordered' node types do not contain a PRB field...
+            return document.containsField(PRB) ?  new BucketedChildReferences(document, this) : ImmutableChildReferences.EMPTY_CHILD_REFERENCES;
```
rhauch: Is this check to protect against bugs? The comment is confusing -- if there is no `PRB` field in the document then it returns an empty `ChildReferences`.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/SessionNode.java
```diff
@@ -1131,14 +1131,20 @@ public ChildReferences getChildReferences( NodeCache cache ) {
...
-        return nodeTypes == null || nodeTypes.allowsNameSiblings(primaryType, mixinTypes);
+        if (nodeTypes == null) {
+            return true;
+        }
+        Name primaryType = getPrimaryType(cache);
+        if (nodeTypes.isUnorderedCollection(primaryType)) {
+            return false;
+        }
+        Set<Name> mixinTypes = getMixinTypes(cache);
+        return nodeTypes.allowsNameSiblings(primaryType, mixinTypes);
```
rhauch: Another option might have been to change `NodeTypes.allowsNameSiblings` to also check whether unordered, since all unordered collections do not allow SNS. That might be better only if the logic above is repeated elsewhere.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/DocumentTranslator.java
hchiorean: the node types instances (from `jcr:system` area) won't have the `PRB` field but still be unordered collections. This field is added to the document only when nodes with these types are created at runtime.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/SessionNode.java
hchiorean: it's not repeated elsewhere, but I agree with you. `NodeTypes` is probably the better place for this.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/WritableSessionCache.java
```diff
@@ -144,11 +147,19 @@ public WritableSessionCache( ExecutionContext context,
...
     }
 
     protected final void assertInSession( SessionNode node ) {
         assert this.changedNodes.get(node.getKey()) == node : "Node " + node.getKey() + " is not in this session";
     }
+    
+    protected NodeTypes nodeTypes() {
+        if (this.nodeTypes == null && workspaceCache().repositoryEnvironment() != null) {
+            this.nodeTypes = workspaceCache().repositoryEnvironment().nodeTypes();
+        }
```
rhauch: This class is `@ThreadSafe` and the `nodeTypes` field is not marked as `volatile`. However, I *think* that might be okay since the actual `NodeTypes` reference returned by `workspaceCache().repositoryEnvironment().nodeTypes()` should always be the same instance. In that case, this logic is idempotent, and its okay for the field to not be `volatile`.

hchiorean: it's always the same instance and it will always be the repository's node types.
the problem is that this field may be null while the repository is initializing (i.e. during the RunningState ctr) while also `session.save` calls are being performed.
So after a repository has started up this should never be null.

modeshape-jcr/src/main/resources/org/modeshape/jcr/modeshape_builtins.cnd
```diff
@@ -140,6 +140,22 @@
     < 'STRING', 'URI', 'BINARY', 'LONG', 'DOUBLE', 'DECIMAL', 'BOOLEAN', 
       'DATE', 'NAME', 'PATH', 'REFERENCE', 'WEAKREFERENCE', 'UNDEFINED'
 
+// -----------------------
+// Large unordered collections
+// -----------------------
+// Abstract base type for all unordered collections.
+
+[mode:unorderedCollection] abstract
+  + * (nt:base) = nt:unstructured abort
```
rhauch: At present, the `mode:unorderedCollection` is an abstract node type that allows any number of child nodes of any node type. Admittedly I designed it via this [MODE-2109 comment](https://issues.jboss.org/browse/MODE-2109?focusedCommentId=12928688&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-12928688). But doesn't this limit a user's ability to design their own structure? For example, if a user has a node type `acme:foo` that contains large numbers of `acme:bar` nodes, then the `acme:foo` must extend `mode:unorderedCollection` and thus would technically allow any numbers of non-`acme:bar` nodes, too.

Would it be better for `mode:unorderedCollection` to be mixins instead, with no restriction on the child node definitions other than they cannot allow SNS? The `NodeTypes` class could certainly enforce this and any other restrictions (like no versioning, etc.). And, I'm not sure much (if any) of the logic will need to change (other than any code that explicitly sets this), since a node's primary type or mixin types might ultimately extend `mode:unorderedCollection` the same way if it were a node type or a mixin type.

What do you think?

rhauch:
Apart from the above questions, this is an excellent implementation of the original concept.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/DocumentTranslator.java
rhauch: So is it possible for an `EmptyCollection` to be returned here when it is in fact not empty?!?

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/WritableSessionCache.java
rhauch: As long as this method is not called during startup, then this will always return non-null. I'm questioning whether the `nodeTypes` field should be marked as `volatile` to ensure it is readable from multiple threads. I think it does not need to be marked as `volatile`: if one thread calls this method and sets the non-volatile `nodeTypes` field, another thread (shortly after that) might call the method and see a null `nodeTypes` field, and that second thread will just reset the `nodeTypes` field to the same reference. IOW, I think it is okay and I'm just looking for confirmation and a comment about the logic being idempotent.

modeshape-jcr/src/main/resources/org/modeshape/jcr/modeshape_builtins.cnd
hchiorean: I don't think mixins are a good idea because they are dynamic - they can be added/removed at runtime. Supporting this - changing how children are stored for nodes - would be highly complex.

I think the original idea of having 4 base types: `mode:unorderedXXXCollection` is a lot better since that means node types will stay immutable. Users are expected to extend any of the 4 types based on their scenario.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/WritableSessionCache.java
hchiorean: i was just explaining why the field is non final (which would solve the concurrency issue).
Other than that it is idempotent (instance wise), so it should not affect the `Threadsafe` contract.

modeshape-jcr/src/main/resources/org/modeshape/jcr/modeshape_builtins.cnd
rhauch: What if we were to require the **primary type** have `mode:unorderedCollection` as a base? That might even simplify the code a bit, since we'd never have to check mixins.

Anyway, isn't it possible to change the primary type of an item?

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/WritableSessionCache.java
rhauch: Right. But that's why I'm suggesting a brief comment, like `// this is idempotent` so that someone reviewing the code in a year can easily see that this logic still is `@ThreadSafe`.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/DocumentTranslator.java
hchiorean: no, it is not.
I was trying to explain that the nodes types instances (e.g. `jcr:system/jcr:nodeTypes/mode:unorderedTinyCollection`) is at runtime, an unordered collection instance which will/can never have a `$prb` field because it's the actual node type instance (sort of like the meta-data). 
That's the only time when `isUnorderedCollection && !document.containsField(PRB)` can be `true`

rhauch: Okay, can you add a comment that explains this? The code is not clear without it, and I never would have surmised this is the reason why it might return an `EmptyChildReferences`.

modeshape-jcr/src/main/resources/org/modeshape/jcr/modeshape_builtins.cnd
hchiorean: >Anyway, isn't it possible to change the primary type of an item?

good point, I missed that. Then that's an extra validation that should be there - i.e. you should not be able to change the node type to/from such a collection.

>What if we were to require the primary type have mode:unorderedCollection as a base? That might even simplify the code a bit, since we'd never have to check mixins.

not sure I follow: we don't test mixins right now and since `mode:unorderedCollection` is abstract, one of the 4 subtypes should be used - which also defines the number of buckets.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/WritableSessionCache.java
hchiorean: ok

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/DocumentTranslator.java
hchiorean: the comment is 
> // the base 'mode:unordered' node types do not contain a PRB field...

but I can add more detail :)

rhauch: > I was trying to explain that the nodes types instances (e.g. `jcr:system/jcr:nodeTypes/mode:unorderedTinyCollection`) is at runtime, an unordered collection instance which will/can never have a `$prb` field because it's the actual node type instance (sort of like the meta-data). 

Really? Isn't every child node under `jcr:system/jcr:nodeTypes` an instance of `jcr:nodeType` (or something like that), which does not extend `mode:unorderedCollection`? Or, does this happen when `nodeTypes` is null? If that is the case, then I might change the logic to be something like:

        if (isUnorderedCollection) {
            if ( document.containsField(PRB) ) {
                return new BucketedChildReferences(document, this);
            }
            // Otherwise, we're in startup mode ...
            assert nodeTypes == null;
            return ImmutableChildReferences.EMPTY_CHILD_REFERENCES; 
        }

> but I can add more detail :)

Yes, please add more detail. :-) I had no idea what this code or comment was trying to signify.

modeshape-jcr/src/main/resources/org/modeshape/jcr/modeshape_builtins.cnd
hchiorean: I guess we can use mixins, as long as we enforce constraints like: 
a) no unordered collection mixin can be added or removed if the node on which it is added/removed has any children.
b) if a node already has such a mixin, a second one is not allowed
That means we could a) remove the base `mode:unorderedCollection` type and b) transform the 4 subtypes into mixins.
@rhauch thoughts ?

rhauch: Yes, I agree. Therefore, it's possible to:

* dynamically add/remove the "unorderedness" to any node with 0 children; or
* statically define a node type (regular or mixin) as being "unordered" by extending one of the unordered mixins (and using that as a primary type for nodes).

Fantastic.

hchiorean:
added an additional commit which changes the node types to mixins plus a number of changes both based on the review and the fact that large collections are now mixins.

hchiorean:
I guess the other option is to change the `NodeTypes` class and perform the `orderable` and `noSNS` checks up-front....

rhauch:
> I guess the other option is to change the NodeTypes class and perform the orderable and noSNS checks up-front....

Yes, I think that'd be best: don't let users create unusable node types.

hchiorean:
amended the previous commit, adding up-front validation of all unordered collection node types and changing the types of nodes used in testing.

rhauch:
Looks great! Very nice work on this complicated feature.

awoods:
This is wonderful. Thank you, @hchiorean!
My one concern is around your initial comment on versioning. Is it true that unordered-collections are mutually exclusive with versioning? That is a hard trade-off to make.

hchiorean:
> Is it true that unordered-collections are mutually exclusive with versioning? 

That's correct. Unordered collections cannot be versioned because it's simply too complex to handle. Even if they were, it would literally "blow up" the versions store because those nodes & hierarchies are not unordered themselves.

>That is a hard trade-off to make.

MODE-2109 is a feature - a nice to have - which was complex enough to implement as it is. We always welcome contributions so if you want to try and look at how versioning would fit into this, that's perfectly fine. Apart from that, implementation wise, this is as far as MODE-2109 will go.

scossu:
Thank you very much for this fix, @rhauch, @hchiorean. Some documentation about how this works and caveats/limitations would be extremely helpful after this is released. 

I have a couple of questions in the meantime: 

- Will performance degrade sensibly for nodes with more than 65K children? In other words, would you still advise against putting millions of nodes under the root node? 
-  Can this feature be applied on a per-node basis in order to take advantage of versioning for nodes that don't have many children? 

Thanks.

rhauch:
Until it can be documented, be sure to see https://issues.jboss.org/browse/MODE-2109 for more detail and background.

PR #1463, opened 2015-09-03
MODE-2478 Updates ISPN to 7.2.4.Final

hchiorean:
This fixes the problem of the marshalling artifact which is now contained in the ISPN AS artifact.


rhauch:
Simple enough.

PR #1464, opened 2015-09-08
MODE-1727 Allows the possibility of setting a custom Locale instance when querying

hchiorean:
This will be used in turn when performing string comparisons.


rhauch:
Looks good.

PR #1465, opened 2015-09-15
MODE-2502 Adds retrying logic when a node cannot locate its parent self-reference

hchiorean:
The code will now attempt to retry loading the parent and its references after removing it from the workspace cache first. This ensures that a "fresh" copy of the parent is always loaded from the persistent store.


rhauch:
Looks good. Glad it was pretty simple and straightforward.

PR #1466, opened 2015-09-22
MODE-1903 Adds incremental reindexing capabilities to the repository

hchiorean:
The incremental reindexing comes in 2 flavors and only works if the repository journal is enabled:

* via the Workspace API:
```java
void reindexSince( long timestamp ) throws RepositoryException;
Future<Boolean> reindexSinceAsync( long timestamp ) throws RepositoryException;
```
* via a new reindexing configuration:
```javascript
"reindexing" : {
        "async" : true | false, //default true
        "mode" : "incremental" | "if_missing", //default if_missing
    }
```
In both cases, the code works similarly: the repository will look at the journal and will get only the `NodeKeys` which have changed since the given timestamp. Then, based on the *current* repository data for those node keys, the indexes will be updated (if the nodes are still present) or removed (if the nodes have been removed)

The index providers have the capability, via a `protected` method, to tell the repository what the latest successful index update time is. If they offer this information instead of `null`, the repository will use this and perform `incremental` reindexing if configured to do so. Otherwise, if `incremental` reindexing is configured, the repository will log a warning at startup.

The above means that *each index provider is responsible for storing the latest successful index update time*, if it wants to support incremental reindexing. In the case of the `LocalIndexProvider`, this is done via simple additional field stored in the `MapDB` file.

This PR does not add support for the new reindexing configuration to the AS kit, this will come via a separate PR.


rhauch:
Does it consider a node that has been deleted as having been "changed"?

hchiorean:
a deleted node will be removed from the indexes. In other words, its NodeKey will appear in the ChangeSets from the journal, but it will not be present in the repository anymore. at that point the RepositoryQueryManager will tell the IndexWriter to remove the stored values for that node.

modeshape-jcr/src/main/java/org/modeshape/jcr/RepositoryQueryManager.java
```diff
@@ -249,16 +256,88 @@ protected final QueryEngine queryEngine() {
...
+            }
+            incrementalIndexingProviders.add(provider);
+            earliestTimestamp = Math.min(earliestTimestamp, latestIndexUpdateTime);
+        }
+        if (incrementalIndexingProviders.isEmpty()) {
+            // none of the providers support incremental reindexing so nothing to do...
+            return;
+        }
+        assert earliestTimestamp != Long.MAX_VALUE;
+        Set<NodeKey> changedNodes = journal.changedNodesSince(earliestTimestamp);
```
rhauch: Isn't it pretty likely that the set of changed node keys might be very large, especially when adding a new process to the cluster? And might that cause memory issues?

I know this is trying to save the amount of work that has to be done by computing the set of changed node keys, but wouldn't it also be possible for the journal to efficiently return an `Iterator<NodeKey>` and then for the `reindexSince(...)` method to use the iterator? The journal could back this by a `Set`, in which case the current behavior is achieved, but it could also back it by an iterator that literally walks the journal and possibly returns a node more than once. The former is much easier to start out with, but the latter is much more scalable for situations where a lot of nodes (perhaps all of them) have been changed. Changing the signature would give us flexibility, while the `reindex` method used on line 326 wouldn't really care.

modeshape-jcr/src/main/java/org/modeshape/jcr/RepositoryQueryManager.java
```diff
@@ -415,6 +499,69 @@ public void reindexContent( JcrWorkspace workspace,
             reindexContent(workspaceName, cache, node, depth, path.isRoot(), getIndexWriter());
         }
     }
+    
+    protected void reindexSince( JcrWorkspace workspace,
+                                 long timestamp ) {
+        ChangeJournal journal = runningState.journal();
+        assert journal != null;
+        Set<NodeKey> changedNodes = journal.changedNodesSince(timestamp);
+        if (changedNodes.isEmpty()) {
```
rhauch: Same comment above about using `Iterable<NodeKey>` here.

modeshape-jcr/src/main/java/org/modeshape/jcr/RepositoryQueryManager.java
```diff
@@ -415,6 +499,69 @@ public void reindexContent( JcrWorkspace workspace,
...
+                writer.remove(workspaceName, nodeKey);
+            }
+        }   
+    }
+
+    protected Future<Boolean> reindexSinceAsync( final JcrWorkspace workspace,
+                                                 final long timestamp ) {
+        ChangeJournal journal = runningState.journal();
+        assert journal != null;
+        Set<NodeKey> changedNodes = journal.changedNodesSince(timestamp);
```
rhauch: ... and here, too. :-)

modeshape-jcr/src/main/java/org/modeshape/jcr/journal/LocalJournal.java
```diff
@@ -265,6 +258,19 @@ public Records recordsNewerThan( DateTime changeSetTime,
...
+    public Set<NodeKey> changedNodesSince( long timestamp ) {
+        Records records = recordsNewerThan(new DateTime(timestamp), true, false);
+        if (records.isEmpty()) {
+            return Collections.emptySet();
+        }
+        Set<NodeKey> results = new LinkedHashSet<>();
+        for (JournalRecord record : records) {
+            results.addAll(record.getChangeSet().changedNodes());
+        }
+        return results;
```
rhauch: This seems like it'd be pretty straightforward to return `Iterable<NodeKey>` here that would return a custom `Iterator` that would walk the records during `hasNext()` and/or `next()`.

rhauch:
This is excellent work and a really great enhancement that is implemented very nicely, @hchiorean. My only suggestion is described above and would make things a bit more flexible and eliminate a potential out-of-memory exception during reindexing.

hchiorean:
@rhauch: good point about the iteration, I will make the change. 

The second thing I wanted to ask you was about naming: the current (default) reindexing behavior I've named it `full`, but the semantic is really: 
>will perform full repository reindexing if there is at least one index (belonging to any provider) which is out of date/new". 

So the question is: should the naming be `full` or `if_missing` ? The latter is what we've used in 3.x, but I've always been confused by it. Is there a better one perhaps ?

rhauch:
Yeah, the naming is kind of tough. Since the context of the mode is within `reindexing`, `if_missing` is a bit more descriptive to me, though `auto` might not be horrible either. Using `full` makes it seem to me that upon restart the repository will do a complete reindexing, and that's not accurate.

hchiorean:
ok, sounds like `if_missing` is the better choice.

hchiorean:
Added an additional commit which:
* addresses the iteration issue (will now expose an `Iterator` instance)
* uses a fixed MapDB key serializer
* upgrades MapDB to `1.0.8`
* adds a validation to `TimeBaseKeys`
* renames the `full` reindexing mode to `if_missing`

rhauch:
Looks great.

PR #1469, opened 2015-10-01
MODE-2159 Changes the default indexing SPI adding default implementations for a lot of CRUD operations

hchiorean:
These changes come the fact that both the Local index provider and the Lucene index provider have very similar logic in a lot of places involving CRUD operations on indexes starting from index definitions. Therefore, this commit "abstracts"  a lot of the logic of the Local index provider and makes it default behavior.

In addition, this commit also adds generic support for multi-column indexes in the form of composite `IndexChangeAdapters`.

This doesn't change any of the current implementation logic (MapDB-wise) behind the local indexes, so it doesn't break backwards compatibility.


hchiorean:
@rhauch: I created this PR because these design changes are precursor to the actual Lucene index implementation (and most likely other implementations as well) as they should substantially simply the CRUD & index-planning part of the effort.

modeshape-jcr/src/main/java/org/modeshape/jcr/spi/index/provider/IndexChangeAdapters.java
```diff
@@ -334,26 +439,35 @@ protected void removeNode( String workspaceName,
...
+                                    NodeKey key, 
+                                    Name primaryType, 
+                                    Set<Name> mixinTypes, 
+                                    NodeKey parent,
+                                    Path newPath, 
+                                    Path oldPath, 
+                                    Path reorderedBeforePath ) {
+            String nodeKey = nodeKey(key);
+            index().remove(nodeKey, valueOf(oldPath));
+            index().add(nodeKey, valueOf(newPath));
```
rhauch: How does this change any of the indexed information for other nodes that appear *after* the `reorderedBeforePath`?

modeshape-jcr/src/main/java/org/modeshape/jcr/spi/index/provider/IndexChangeAdapters.java
```diff
@@ -393,14 +507,20 @@ protected Name convert( Path path ) {
         protected Name convertRoot( Path path ) {
             return Path.ROOT_NAME;
         }
+
+        @Override
+        protected void reorderNode( String workspaceName, NodeKey key, Name primaryType, Set<Name> mixinTypes, NodeKey parent,
+                                    Path newPath, Path oldPath, Path reorderedBeforePath ) {
+            // reordering should not really change the name...
```
rhauch: Might it not change the SNS of the node?

modeshape-jcr/src/main/java/org/modeshape/jcr/spi/index/provider/IndexChangeAdapters.java
```diff
@@ -412,14 +532,20 @@ protected String convert( Path path ) {
         protected String convertRoot( Path path ) {
             return "";
         }
+
+        @Override
+        protected void reorderNode( String workspaceName, NodeKey key, Name primaryType, Set<Name> mixinTypes, NodeKey parent,
+                                    Path newPath, Path oldPath, Path reorderedBeforePath ) {
+            // reordering should not really change the name...
```
rhauch: This is true for local names.

modeshape-jcr/src/main/java/org/modeshape/jcr/spi/index/provider/IndexProvider.java
```diff
@@ -630,11 +739,82 @@ private boolean matchesType( IndexDefinition defn,
...
+     * @param calculator the calculator that should be used to record plan information for the index; never null
+     * @param defn the definition for the index; never null
+     * @return an {@link IndexUsage} instance, never null.
+     * 
+     * @see #planUseOfIndex(QueryContext, IndexCostCalculator, String, ManagedIndex, IndexDefinition) 
+     */
+    protected IndexUsage evaluateUsage( QueryContext context,
+                                        IndexCostCalculator calculator,
+                                        IndexDefinition defn ) {
+        return null;
```
rhauch: Line 795 says this never returns null. Should this be made abstract rather than return an invalid value by default?

modeshape-jcr/src/main/java/org/modeshape/jcr/spi/index/provider/IndexProvider.java
```diff
@@ -836,23 +1044,47 @@ private void addProvidedIndex( ProvidedIndex index ) {
...
+
+            @SuppressWarnings( "synthetic-access" )
+            @Override
+            public void afterIndexing() {
+                managedIndex.enable(true);
+                logger().debug("Enabled index '{0}' from provider '{1}' in workspace '{2}' after reindexing has completed",
+                               defn.getName(), defn.getProviderName(), workspaceName);
+            }
+        });
+    }
```
rhauch: Nice.

rhauch:
So my biggest comment is that this appears this will break existing index providers. Is that correct? If so, is that something we can/want to do in a minor release?

hchiorean:
The changes in the `IndexProvider` base class should not break any providers (the added methods are non-abstract and return default values - null) while the previously `abstract` methods are not non-abstract, but still binary-compatible.

The only change which I guess could break existing providers is the fact that `ManagedIndex` changed from being an interface to being an `abstract class` with some functionality.  I guess I could work around this by reverting it back to an interface and adding a new class which implements `ManagedIndex`.

On the other hand, I really want to get these changes in before working on the Lucene provider. Also, if there are other provider implementations out there, they were certainly not even attempted to be contributed to the codebase, so since all the changes are in the `modeshape-jcr` package (not the `-api` package) I don't think the change to `MangedIndex` is considered that "major".

modeshape-jcr/src/main/java/org/modeshape/jcr/spi/index/provider/IndexChangeAdapters.java
hchiorean: it doesn't. should it ? :)

modeshape-jcr/src/main/java/org/modeshape/jcr/spi/index/provider/IndexChangeAdapters.java
hchiorean: isn't the SNS a function of the path (segment) ? I thought that the node name (or mode:localName) does not reflect the SNS index. If it does however, then I'll have to change this,

rhauch:
I'll leave it up to you whether to change `ManagedIndex` from interface to abstract class, or whether to add a default implementation and keep `ManagedIndex` an interface. Really depends upon how gross the latter would be vs the risk of someone using a custom provider. IIUC, those using the built-in provider will not have a problem, since nothing has changed about the actual indexes (e.g., the new ModeShape version would be able to read the old MapDB files produced by the prior ModeShape version).

modeshape-jcr/src/main/java/org/modeshape/jcr/spi/index/provider/IndexChangeAdapters.java
rhauch: For nodes that do not use SNS, then you are correct in that nothing about child nodes after the `reorderedBeforePath` need be changed.

However, if SNS nodes are being reordered, then reordering changes the SNS index of the node being reordered *and* those nodes after `reorderedBeforePath`. I can give an example if that'd help.

modeshape-jcr/src/main/java/org/modeshape/jcr/spi/index/provider/IndexChangeAdapters.java
rhauch: *sigh*. Yeah, you're right that the name does not include SNS. No worries here.

modeshape-jcr/src/main/java/org/modeshape/jcr/spi/index/provider/IndexChangeAdapters.java
hchiorean: well this sucks because I don't see how this case can be handled: each reordering event contains only information about the node which was reordered - i.e. which node, what was the old path and what is the new path - which is what the index attempts to handle.

however, if this causes the side affect of altering the node paths of other SNS siblings which are also present in the index, there's is no way to make any changes for those as there's simply not enough information to do so.

So unless I'm missing something, I don't see any other options than raise a JIRA and leave it open :) (I'm also pretty sure the same problem exists in ModeShape 3).

hchiorean: well, I guess it's not as bad as I though (ran a quick SNS reordering test). The only actual changes are between the source node and the target node, since SNSs are computed each time (not stored).
So given `node[1], node[2], node[3], node[4]` performing a `reorder(node[4],node[1])` will still yield the same result, but the actual keys which have changed positions are `node[1]` and `node[4]`. 
So there is still some work to update the `reorder` events, but it's doable (unlike my previous "fear")

rhauch: First, I *think* you can use the `oldPath` and `newPath` to check the SNS index of the node that was reordered. If the SNS index is `1` in both places, then that should imply that no other nodes were affected (right?), and that the existing logic is sufficient.

However, I do think we have to address what happens when the SNS indexes are different. But, first we need to make sure that no other events are generated that correspond to the other changes. That would be ideal, because those other events would be handled. (But I don't think there are other events, though it may be easier to generate additional events for those cases. Something to consider.)

But, if the SNS indexes are different and no other events handle the other affected nodes, then we probably can rely upon the index provider to help. For example, it can search its index to find all nodes that have the same path (except different SNS indexes). The question is whether to add a method that does this, or to use queries to do this. Either way, this change adapter can then use the results to tell the index to change the other affected nodes. 

And of course, if we add a method on the index provider and we want to keep it backward compatible, we'd want to have a default implementation. If the default implementation did nothing (essentially saying no other nodes are affected by this SNS change), then the index could become inconsistent. However, if the default implementation could return a result that said it doesn't know (e.g., null rather than a collection or iterator), then can the change adapter cause a reindexing of just the children of the parent (and not descendants of the children)? That's not ideal, but at least it would keep the index from becoming inconsistent; and once a provider implements the method correctly, then it becomes efficient again.

WDYT?

rhauch: I just read your previous comment, and I'm glad it (may) not be that difficult.

hchiorean:
amended the original commit, bringing back the `ManagedIndex` interface and added a new commit which fixes the reordering issue by adding more information to the event.

hchiorean:
well that's a good point. This is indeed as complex as I originally thought. Moreover, I'm highly reluctant to change the SPI of the index providers atm to handle this edge case.

So, considering that:

a) this is not handled in any way in ModeShape 3 (which stores the path in the `NodeInfo` objects and in the case of reorderings simply reindexes the node which was reordered without any other changes)
b)  was not handled at all in ModeShape 4 in  the previous version of the Local Indexes (the `nodeReordered` events were not processed at all)
c) no bugs have been raised in the community on this until now

I'm proposing that the `path` based indexes simply ignore reorderings of SNS. I'll raise a JIRA for this and sometime in the future this can be handled. 

Thoughts ?

rhauch:
Sounds good.

hchiorean:
removed the commit which added the `NodeKey` information of the reordered before node and added a TODO in the `PathBasedChangeAdapter` which points to https://issues.jboss.org/browse/MODE-2510

rhauch:
Looks good.

PR #1470, opened 2015-10-13
MODE-2159 Updates the index adapters SPI adding a property name for each add/remove operation

hchiorean:
Index providers are free to ignore the property if their logic does not require it (for example the Local Index provider) but other (e.g. Lucene) are interested in the property name so they can add it to the indexing information. 

Also, this PR changes the default index planning logic adding the ability for Index Providers to process multiple constraints, not just one like the `LocalIndexProvider` did.


rhauch:
Looks good.

okulikov:
Looks good.

PR #1471, opened 2015-10-20
MODE-2159 Adds support for storing indexes in Lucene

hchiorean:
This uses the latest version of Lucene `5.3.1`. It supports both a simple (default) configuration:
```javascript
"indexProviders" : {
        "lucene" : {
            "classname" : "lucene",
            "directory" : "target/LuceneIndexProviderQueryTest"
        }
    }
```
and a more advanced one:
```javascript
"indexProviders" : {
        "lucene" : {
            "classname" : "lucene",
            "lockFactoryClass" : "org.apache.lucene.store.NoLockFactory",
            "directoryClass" : "org.apache.lucene.store.RAMDirectory",
            "analyzerClass" : "org.apache.lucene.analysis.ro.RomanianAnalyzer",
            "codec" : "Lucene53"
        }
    }
```
It supports all 4 index types, including text indexes which will only be used for FTS queries. Text indexes will not be selected *for anything else*. `LIKE` is supported by the default `value` indexes. If a FTS query doesn't contain a property name (uses `*`), __all FTS indexes which match the selector/node types__ will apply/be asked to contribute to node filtering.

It supports both single and multi-column indexes, with the caveat that multi-column indexes __will perform worse__ than single column. This is something which __will be documented__ and has to do with the fact that Lucene *does not support document merging (updating)*. This means that any changes to a multi-column index require an additional lookup to check if a document already exists and if so, creating the new "merged" document and then issuing an `update` to Lucene which will remove the old document and add the new one.

_This PR does not include AS support for the new index provider, which will be added via another PR_

### DIFF
--- a/boms/modeshape-bom-embedded/pom.xml
+++ b/boms/modeshape-bom-embedded/pom.xml
@@ -80,6 +80,7 @@
         <poi.version>3.11-beta2</poi.version>
         <jcommander.version>1.5</jcommander.version>
         <wsdl4j.version>1.6.2</wsdl4j.version>
+        <version.org.apache.lucene>5.3.1</version.org.apache.lucene>
         <!-- ***************** -->
         <!-- Repository Deployment URLs -->
         <!-- ***************** -->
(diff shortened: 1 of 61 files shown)


rhauch:
Looks good.

PR #1475, opened 2015-10-27
MODE-2514 Fixes the FS connector's handling of root folders which don't match a configured filter

hchiorean:


rhauch:
Looks good.

PR #1479, opened 2015-11-06
MODE-2516 Updates JDK to 1.8 and jboss-parent to the latest version (19)

hchiorean:
This is the first significant commit of the `5.x` series and contains a number of significant changes:
- the naming of Maven version properties changed to adopt the standard pattern: `version.groupId.artifactId`
- build system and dependency updates so that the latest Maven plugin versions function correctly
- __removing Joda-Time__ library and changing the code to use the new JDK 1.8 date-time API
- updating source code to avoid compiler and javadoc warnings
- a fix for [MODE-2520](https://issues.jboss.org/browse/MODE-2520)


rhauch:
So was the fix for MODE-2520 dependent upon Java 8? It's not clear why that was included in this PR (not even as a separate commit).

modeshape-jcr-api/src/main/java/org/modeshape/jcr/api/value/DateTime.java
```diff
@@ -274,151 +187,21 @@
...
-     */
-    DateTime plusWeeks( int weeks );
-
-    /**
-     * Add the specified number of years from this time instant.
-     * 
-     * @param years the number of years to add
-     * @return the instance in time the specified number of years after this instant
-     */
-    DateTime plusYears( int years );
```
rhauch: Should we handle changing public APIs like this as separate issues? Besides, couldn't all of these be implemented by default by using `toZonedDateTime` (or `toLocalDateTime`)?

modeshape-jcr/src/main/java/org/modeshape/jcr/value/basic/ModeShapeDateTime.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    @Override
+    public ZonedDateTime toZonedDateTime() {
+        return instance;
+    }
+
+    @Override
+    public LocalDateTime toLocalDateTime() {
+        return instance.toLocalDateTime();
```
rhauch: You could move any of these methods that use `instance` to be default methods on the interface that use `toZonedDateTime()`. It's certainly not wrong to do it like you did, especially since there is no other impl, and because there's not much point in doing this _for the sake of moving to Java 8_. But as a general rule default methods are more idiomatic in Java 8, and it really does help to keep impls simple.

modeshape-jcr-api/src/main/java/org/modeshape/jcr/api/value/DateTime.java
hchiorean: >Should we handle changing public APIs like this as separate issues?

Ideally yes. But in practice after changing the JDK and jboss-parent I started fixing issues incrementally, so it was really had to split each change. I _could_ try to do it, but it would be probably hard since there are a lot of intersecting issues.
For example: fixing javadoc and compiler warnings, code style refactorings and actual changing code may affect the same file.

> Besides, couldn't all of these be implemented as default methods by using toZonedDateTime (or toLocalDateTime)?

Yes, but these particular methods were already used throughout the existing code, so keeping those methods as they were was easier than changing a lot of other places.

hchiorean:
>So was the fix for MODE-2520 dependent upon Java 8? It's not clear why that was included in this PR (not even as a separate commit).

MODE-2520 was the result of some Lucene unit tests failing all-of-a-sudden on JDK 1.8. The exact same tests were working perfectly on JDK 1.7. I don't know exactly why that is, but upon closer inspection is turned out that the indexing code had a legit bug that needed fixing.

modeshape-jcr-api/src/main/java/org/modeshape/jcr/api/value/DateTime.java
rhauch: Actually, I was talking about keeping the methods that were removed above, and implementing them in terms of ZonedDateTime. Seems especially true if the methods were already used throughout the existing code.

rhauch:
> MODE-2520 was the result of some Lucene unit tests failing all-of-a-sudden on JDK 1.8. The exact same tests were working perfectly on JDK 1.7. I don't know exactly why that is, but upon closer inspection it turned out that the indexing code had a legit bug that needed fixing.

Sure, but I guess I would have made a separate commit. Often when I have a bunch of local changes (when things snowball), I'll stage them manually (sometimes even by blocks) so that I can group the changes into meaningful commits. IMO, large commits that change lots of seemingly unrelated things make things messy.

That's the only comment I have about this PR. Otherwise, technically, all the changes look fine.

modeshape-jcr/src/main/java/org/modeshape/jcr/value/basic/ModeShapeDateTime.java
hchiorean: agreed; if you prefer I can change them to be `default` methods. 

In this particular case however, it _feels_ like changing them to be `default` methods somehow exposes an internal implementation detail, which makes me _feel_ like the `interface` is misused in that it's turning out to be more as an `abstract class`

hchiorean:
@rhauch: ok, I'll go back through all the changes and I'll try to split this commit into 3: one with the build & JDK changes, one for MODE-2520 and one for the date-time changes. Especially the last one might be tricky, but we'll see.

modeshape-jcr/src/main/java/org/modeshape/jcr/value/basic/ModeShapeDateTime.java
rhauch: I agree that it is questionable in cases like this, where the interface is intended to be a public API. So lots of negatives to making them default methods. However, sometimes it is still useful to put default methods in a public API when their implementation adds value (e.g., "oh, that makes sense"). 

Overall, changing these to default methods is probably less useful, so I think it's perfectly fine to keep them as is.

modeshape-jcr-api/src/main/java/org/modeshape/jcr/api/value/DateTime.java
hchiorean: actually, 99% percent of the removed methods were unused. the 1% that was used and was removed was a really quick change in one or two places.

I left those `plus` methods alone however because the default JDK 1.8 API does not support adding/subtracting any duration unit longer than a `day` so these methods are actually an "enhancement" which could not be expressed as-is with the default API.

rhauch: Perhaps they were not used in our codebase, but isn't the interface part of ModeShape's public API? Now, changing the public API and removing methods is fine (especially since this is a major release), but IMO breaking changes for users should a) have a dedicated issue, and b) ideally have a separate commit.

hchiorean: I've opened https://issues.jboss.org/browse/MODE-2522

hchiorean:
amended the PR, splitting it into 4 commits: the 3 already mentioned in my previous comment, plus an additional one which removes deprecated API.

rhauch:
Looks great.

PR #1481, opened 2015-11-12
MODE-2526 Replaces the workspace cache with a simple LRU cache based on a ConcurrentLinkedMap

hchiorean:
This is also updates both the JSON configuration and the Wildfly configuration with a simple setting - `cacheSize` - which controls the maximum number of elements from the workspace cache.

The cache implementation is the one from https://github.com/ben-manes/concurrentlinkedhashmap/wiki/Design, which is simple and lightweight enough for what we need (plus ASL licensed)


deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/ModeShapeSubsystemXMLReader_3_0.java
```diff
@@ -31,7 +31,7 @@
 import org.jboss.staxmapper.XMLElementReader;
 import org.jboss.staxmapper.XMLExtendedStreamReader;
 
-public class ModeShapeSubsystemXMLReader_2_1 implements XMLStreamConstants, XMLElementReader<List<ModelNode>> {
+public class ModeShapeSubsystemXMLReader_3_0 implements XMLStreamConstants, XMLElementReader<List<ModelNode>> {
```
rhauch: Will we need to keep a `ModeShapeSubsystemXMLReader_2_1` implementation around that reads the older configuration but builds the newer model? IIUC, that would allow us to parse 4.x configurations, and Wildfly would automatically write them out using the latest XML representation. (I know this is unrelated to this issue, but I would think that if we're going to want one then it'd be easier to start with it sooner than later.) Thoughts?

hchiorean: No it would not, but the only required change would be:
```
 <subsystem xmlns="urn:jboss:domain:modeshape:3.0">
```
So only the namespace number has to be changed, which is more than reasonable IMO. It reflects the change of major version, which will translate into a whole bunch of configuration changes.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/WorkspaceCache.java
```diff
@@ -249,10 +248,8 @@ public CachedNode getNode( NodeKey key ) {
                 CachedNode newNode = new LazyCachedNode(key, doc);
                 try {
                     Integer cacheTtlSeconds = translator().getCacheTtlSeconds(doc);
-                    if (cacheTtlSeconds != null && nodesByKey instanceof org.infinispan.commons.api.BasicCache) {
-                        node = ((org.infinispan.commons.api.BasicCache<NodeKey, CachedNode>)nodesByKey).putIfAbsent(key, newNode,
-                                                                                         cacheTtlSeconds.longValue(),
-                                                                                         TimeUnit.SECONDS);
+                    if (cacheTtlSeconds != null) {
+                        //TODO author=Horia Chiorean date=10/11/2015 description=Cache TTL seconds are not supported by the current cache
```
rhauch: Shouldn't we at least put the node in the cache even if there is a TTL?

hchiorean: As per https://issues.jboss.org/browse/MODE-2529 TTL will be removed altogether and replaced by a boolean which has the semantic of either caching something or not caching it at all.

rhauch: Then wouldn't it be better in this method to completely ignore the `cacheTtlSeconds` value? The current code doesn't ignore it, but when it is specified will _not_ insert something into the cache.

deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/ModeShapeSubsystemXMLReader_3_0.java
rhauch: But when a dev changes the subsystem to use "3.0", won't they have to manually convert their configuration? Is there no automated way to do that?

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/WorkspaceCache.java
rhauch: Okay, I've read your description in https://issues.jboss.org/browse/MODE-2529 and understand the TODO.

rhauch:
Looks good.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/WorkspaceCache.java
hchiorean: The PR for https://issues.jboss.org/browse/MODE-2529  will change the code altogether in this area. The PR and issues for {{5.x}} are not completely independent atm. but I'm trying to create granular PRs to phase out the work. 

This and the upcoming PRs which are subtasks of https://issues.jboss.org/browse/MODE-2525 should not be taken 100% independent of each other really

deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/ModeShapeSubsystemXMLReader_3_0.java
hchiorean: not that i'm aware of, no.

For comparison, https://issues.jboss.org/browse/MODE-2454 changed the version from `2.0` to `2.1` between 2 `4.x` _minor_ releases. This is for a major release and is exactly the same thing....

hchiorean:
@rhauch: so are you ok with merging this as-is ?

rhauch:
@hchiorean, yes, it looks good as-is.

PR #1482, opened 2015-11-13
MODE-2529 Removes the "cacheTtlSeconds" setting and replaces it with a "cacheable" flag

hchiorean:
This also cleans up some of the code for https://issues.jboss.org/browse/MODE-1712 as most of those abstractions weren't really required.


rhauch:
Looks good.

PR #1483, opened 2015-11-16
MODE-2530 Fixes the potential corruption of the version history structure for a versionable node under concurrent access

hchiorean:
The fix is making sure that the version history of a node is only initialized (i.e. substructure created) __after__ the root `jcr:versionStorage` parent has been locked as part of a transaction.


rhauch:
Looks good.

PR #1487, opened 2015-11-26
MODE-2536, MODE-2537 Ads the ability for ModeShape to integrate with JTA and fixes some thread pool issues

hchiorean:
- `MODE-2536` is required because in the absence of ISPN, ModeShape will have to be able to support both local and user transactions, regardless of the environment it's being run in. This commit also removes the `transactionMode` configuration option, as ModeShape should be able to auto-detect if it's being run or not in the context of a user transaction.
- `MODE-2537` fixes the improper shutdown of certain thread pools in the case of a repository crash

### DIFF
--- a/boms/modeshape-bom-embedded/pom.xml
+++ b/boms/modeshape-bom-embedded/pom.xml
@@ -50,6 +50,7 @@
 
     <properties>
         <version.javax.jcr>2.0</version.javax.jcr>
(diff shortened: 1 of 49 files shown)


rhauch:
Outstanding.

hchiorean:
amended the commit based on the above suggestions

PR #1488, opened 2015-11-26
MODE-2162: Store indexes in ElasticSearch

okulikov:
This pull request implements Elasticsearch index provider

### DIFF
--- a/index-providers/modeshape-elasticsearch-index-provider/pom.xml
+++ b/index-providers/modeshape-elasticsearch-index-provider/pom.xml
@@ -1,0 +1,32 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.modeshape</groupId>
+        <artifactId>modeshape-index-providers</artifactId>
+        <version>5.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <!-- The groupId and version values are inherited from parent -->
+    <artifactId>modeshape-elasticsearch-index-provider</artifactId>
+    <packaging>jar</packaging>
+    <name>ModeShape Elasticsearch Index Provider</name>
+    <description>ModeShape index provider that uses Elasticsearch engine</description>
+    <url>http://www.modeshape.org</url>
+    <dependencies>
+        <dependency>
+            <groupId>org.elasticsearch</groupId>
+            <artifactId>elasticsearch</artifactId>
+            <scope>test</scope>
+        </dependency>  
+        <dependency>
+            <groupId>net.java.dev.jna</groupId>
+            <artifactId>jna</artifactId>
+            <scope>test</scope>
+        </dependency>                
+        <dependency>
+            <groupId>org.modeshape</groupId>
+            <artifactId>modeshape-schematic</artifactId>
+        </dependency>
+    </dependencies>      
+</project>
--- a/index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndex.java
+++ b/index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndex.java
@@ -1,0 +1,303 @@
+/*
+ * ModeShape (http://www.modeshape.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.modeshape.jcr.index.elasticsearch;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import javax.jcr.query.qom.Constraint;
+import org.modeshape.common.util.CheckArg;
+import org.modeshape.jcr.ExecutionContext;
+import org.modeshape.jcr.api.index.IndexDefinition;
+import org.modeshape.jcr.index.elasticsearch.client.EsClient;
+import org.modeshape.jcr.index.elasticsearch.client.EsRequest;
+import org.modeshape.jcr.spi.index.IndexConstraints;
+import org.modeshape.jcr.spi.index.provider.ProvidedIndex;
+
+/**
+ * Index stored in Elasticsearch.
+ *
+ * @author kulikov
+ */
+public class EsIndex implements ProvidedIndex {
+
+    private final String name;
+    private final String workspace;
+    private final EsIndexColumns columns;
+    private final Operations operations;
+    private final EsClient client;
+    
+    /**
+     * Creates new index.
+     * 
+     * @param client provides access to the elasticsearch functions.
+     * @param context Modeshape execution context.
+     * @param defn index definition.
+     * @param workspace workspace name where this index will be created.
+     */
+    public EsIndex(EsClient client, ExecutionContext context, IndexDefinition defn, String workspace) {
+        this.client = client;
+        this.name = defn.getName();
+        this.workspace = workspace;
+        this.columns = new EsIndexColumns(context, defn);
+        this.operations = new Operations(context.getValueFactories(), columns);
+        this.createIndex();
+    }
+
+    /**
+     * Creates new index.
+     * 
+     * @param client provides access to the elasticsearch functions.
+     * @param columns columns definition.
+     * @param context Modeshape execution context.
+     * @param name the name of the index
+     * @param workspace workspace name where this index will be created.
+     */
+    protected EsIndex(EsClient client, EsIndexColumns columns, ExecutionContext context, String name, String workspace) {
+        this.client = client;
+        this.name = name;
+        this.workspace = workspace;
+        this.columns = columns;
+        this.operations = new Operations(context.getValueFactories(), columns);
+        this.createIndex();
+    }
+
+    /**
+     * Generates local index name.
+     * 
+     * @return 
+     */
+    private String name() {
+        return name.toLowerCase() + "-" + workspace;
+    }
+
+    /**
+     * Executes create index action.
+     */
+    private void createIndex() {
+        try {
+            client.createIndex(name(), workspace, columns.mappings(workspace));
+            client.flush(name());
+        } catch (IOException e) {
+            throw new EsIndexException(e);
+        }
+    }
+
+    @Override
+    public void add(String nodeKey, String propertyName, Object value) {
+        CheckArg.isNotNull(nodeKey, "nodeKey");
+        CheckArg.isNotNull(propertyName, "propertyName");
+        CheckArg.isNotNull(value, "value");
+
+        EsIndexColumn column = columns.column(propertyName);
+        assert column != null : "Unexpected column for the index " + name();
+
+        try {
+            EsRequest doc = findOrCreateDoc(nodeKey);
+            putValue(doc, column, value);
+            client.storeDocument(name(), workspace, nodeKey, doc);
+        } catch (IOException e) {
+            throw new EsIndexException(e);
+        }
+    }
+
+    @Override
+    public void add(String nodeKey, String propertyName, Object[] values) {
+        CheckArg.isNotNull(nodeKey, "nodeKey");
+        CheckArg.isNotNull(propertyName, "propertyName");
+        CheckArg.isNotNull(values, "values");
+
+        EsIndexColumn column = columns.column(propertyName);
+        assert column != null : "Unexpected column for the index " + name();
+
+        try {
+            EsRequest doc = findOrCreateDoc(nodeKey);
+            putValues(doc, column, values);
+            client.storeDocument(name(), workspace, nodeKey, doc);
+        } catch (IOException e) {
+            throw new EsIndexException(e);
+        }
+    }
+
+    @Override
+    public void remove(String nodeKey) {
+        CheckArg.isNotNull(nodeKey, "nodeKey");
+        try {
+            client.deleteDocument(name(), workspace, nodeKey);
+        } catch (IOException e) {
+            throw new EsIndexException(e);
+        }
+    }
+
+    @Override
+    public void remove(String nodeKey, String propertyName, Object value) {
+        CheckArg.isNotNull(nodeKey, "nodeKey");
+        CheckArg.isNotNull(propertyName, "propertyName");
+
+        try {
+            EsRequest doc = find(nodeKey);
+            if (doc == null) {
+                return;
+            }
+            doc.remove(propertyName);            
+            client.storeDocument(name(), workspace, nodeKey, doc);
+        } catch (IOException e) {
+            throw new EsIndexException(e);
+        }
+    }
+
+    @Override
+    public void remove(String nodeKey, String propertyName, Object[] values) {
+        CheckArg.isNotNull(nodeKey, "nodeKey");
+        CheckArg.isNotNull(propertyName, "propertyName");
+
+        try {
+            EsRequest doc = find(nodeKey);
+            doc.remove(propertyName);
+            client.storeDocument(name(), workspace, nodeKey, doc);
+        } catch (Exception e) {
+            throw new EsIndexException(e);
+        }
+    }
+
+    /**
+     * Searches indexed node's properties by node key.
+     * 
+     * @param nodeKey node key being indexed.
+     * @return list of stored properties as json document.
+     * @throws IOException 
+     */
+    private EsRequest find(String nodeKey) throws IOException {
+        return client.getDocument(name(), workspace, nodeKey);
+    }
+
+    /**
+     * Searches indexed node's properties by node key or creates new empty list.
+     * 
+     * @param nodeKey node key being indexed.
+     * @return list of stored properties as json document or empty document 
+     * if not found.
+     * @throws IOException 
+     */
+    private EsRequest findOrCreateDoc(String nodeKey) throws IOException {
+        EsRequest doc = client.getDocument(name(), workspace, nodeKey);
+        return doc != null ? doc : new EsRequest();
+    }
+
+    /**
+     * Appends specified value for the given column and related pseudo columns
+     * into list of properties.
+     * 
+     * @param doc list of properties in json format
+     * @param column colum definition
+     * @param value column's value.
+     */
+    private void putValue(EsRequest doc, EsIndexColumn column, Object value) {
+        Object columnValue = column.columnValue(value);
+        String stringValue = column.stringValue(value);
+        doc.put(column.getName(), columnValue);
+        doc.put(column.getLowerCaseFieldName(), stringValue.toLowerCase());
+        doc.put(column.getUpperCaseFieldName(), stringValue.toUpperCase());
+        doc.put(column.getLengthFieldName(), stringValue.length());
+    }
+
+    /**
+     * Appends specified values for the given column and related pseudo columns
+     * into list of properties.
+     * 
+     * @param doc list of properties in json format
+     * @param column colum definition
+     * @param value column's value.
+     */
+    private void putValues(EsRequest doc, EsIndexColumn column, Object[] value) {
+        Object[] columnValue = column.columnValues(value);
+        int[] ln = new int[columnValue.length];
+        String[] lc = new String[columnValue.length];
+        String[] uc = new String[columnValue.length];
+
+        for (int i = 0; i < columnValue.length; i++) {
+            String stringValue = column.stringValue(columnValue[i]);
+            lc[i] = stringValue.toLowerCase();
+            uc[i] = stringValue.toUpperCase();
+            ln[i] = stringValue.length();
+        }
+
+        doc.put(column.getName(), columnValue);
+        doc.put(column.getLowerCaseFieldName(), lc);
+        doc.put(column.getUpperCaseFieldName(), uc);
+        doc.put(column.getLengthFieldName(), ln);
+    }
+
+    @Override
+    public void commit() {
+        try {
+            client.refresh(name());
+        } catch (IOException e) {
+            throw new EsIndexException(e);
+        }
+    }
+
+    @Override
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public Results filter(IndexConstraints constraints) {
+        EsRequest query = operations.createQuery(constraints.getConstraints(), constraints.getVariables());
+        return new SearchResults(client, name(), workspace, query);
+    }
+
+    @Override
+    public long estimateCardinality(List<Constraint> constraints, Map<String, Object> variables) {
+        EsRequest query = operations.createQuery(constraints, variables);
+        return new SearchResults(client, name(), workspace, query).getCardinality();
+    }
+
+    @Override
+    public long estimateTotalCount() {
+        try {
+            return client.count(name(), workspace);
+        } catch (IOException e) {
+            throw new EsIndexException(e);
+        }
+    }
+
+    @Override
+    public boolean requiresReindexing() {
+        return true;
+    }
+
+    @Override
+    public void clearAllData() {
+        try {
+            client.deleteAll(name(), workspace);
+        } catch (IOException e) {
+            throw new EsIndexException(e);
+        }
+    }
+
+    @Override
+    public void shutdown(boolean destroyed) {
+        if (destroyed) {
+            try {
+                client.deleteIndex(name());
+            } catch (Exception e) {
+                throw new EsIndexException(e);
+            }
+        }
+    }
+}
--- a/index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndexColumn.java
+++ b/index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndexColumn.java
@@ -1,0 +1,204 @@
+/*
+ * ModeShape (http://www.modeshape.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.modeshape.jcr.index.elasticsearch;
+
+import java.util.Collection;
+import org.modeshape.jcr.ExecutionContext;
+import org.modeshape.jcr.api.value.DateTime;
+import org.modeshape.jcr.value.PropertyType;
+import static org.modeshape.jcr.value.PropertyType.BOOLEAN;
+import static org.modeshape.jcr.value.PropertyType.DATE;
+import static org.modeshape.jcr.value.PropertyType.LONG;
+import static org.modeshape.jcr.value.PropertyType.NAME;
+import static org.modeshape.jcr.value.PropertyType.REFERENCE;
+import static org.modeshape.jcr.value.PropertyType.SIMPLEREFERENCE;
+import static org.modeshape.jcr.value.PropertyType.URI;
+import static org.modeshape.jcr.value.PropertyType.WEAKREFERENCE;
+import org.modeshape.jcr.value.ValueFactories;
+
+/**
+ * Elasticsearch field definition.
+ * 
+ * Provides definition for elasticsearch field and implements value 
+ * conversations with accordance of the column type.
+ * 
+ * @author kulikov
+ */
+public class EsIndexColumn {
+    public final static String LENGTH_PREFIX = "length_";
+    public final static String LOWERCASE_PREFIX = "lowercase_";
+    public final static String UPPERCASE_PREFIX = "uppercase_";
+
+    private final String name;
+    private final PropertyType type;
+    private final ValueFactories valueFactories;
+    
+    /**
+     * Creates new field definition.
+     * 
+     * @param context ModeShape execution context
+     * @param name name of this column.
+     * @param type JCR type of this column.
+     */
+    public EsIndexColumn(ExecutionContext context, String name, int type ) {
+        this.valueFactories = context.getValueFactories();
+        this.name = name;
+        this.type = PropertyType.valueFor(type);
+    }
+
+    /**
+     * Gets the name of this column.
+     * 
+     * @return the name of the column.
+     */
+    public String getName() {
+        return name;
+    }
+
+    /**
+     * Gets the name of the field to store length.
+     * 
+     * @return 
+     */
+    public String getLengthFieldName() {
+        return LENGTH_PREFIX + name;
+    }
+    
+    /**
+     * Gets the name of the field used for store lowered case value.
+     * 
+     * @return field name.
+     */
+    public String getLowerCaseFieldName() {
+        return LOWERCASE_PREFIX + name;
+    }
+
+    /**
+     * Gets the name of the field used for store upper case value.
+     * 
+     * @return field name.
+     */
+    public String getUpperCaseFieldName() {
+        return UPPERCASE_PREFIX + name;
+    }
+    
+    /**
+     * Gets type of this column.
+     * 
+     * @return Jcr type of this column.
+     */
+    public PropertyType getType() {
+        return type;
+    }
+
+    /**
+     * Converts representation of the given value using type conversation rules
+     * between JCR type of this column and Elasticsearch core type of this column.
+     * 
+     * @param value value to be converted.
+     * @return converted value.
+     */
+    protected Object columnValue(Object value) {
+        switch (type) {
+            case PATH:
+            case NAME:
+            case STRING:
+            case REFERENCE:
+            case SIMPLEREFERENCE:
+            case WEAKREFERENCE:
+            case URI:
+                return valueFactories.getStringFactory().create(value);
+            case DATE :
+                return ((DateTime) value).getMilliseconds();
+            default :
+                return value;
+        }
+    }
+
+    protected String stringValue(Object value) {
+        return valueFactories.getStringFactory().create(value);
+    }
+    
+    /**
+     * Converts array of values.
+     * @param value
+     * @return 
+     */
+    protected Object[] columnValues(Object[] value) {
+        Object[] res = new Object[value.length];
+        for (int i = 0; i < value.length; i++) {
+            res[i] = columnValue(value[i]);
+        }
+        return res;
+    }
+    
+    /**
+     * Converts given value to the value of JCR type of this column.
+     * 
+     * @param value value to be converted.
+     * @return converted value.
+     */
+    protected Object cast(Object value) {
+        switch (type) {
+            case STRING :
+               return valueFactories.getStringFactory().create(value);
+            case LONG :
+                return valueFactories.getLongFactory().create(value);
+            case NAME :
+                return valueFactories.getNameFactory().create(value);
+            case PATH :
+                return valueFactories.getPathFactory().create(value);
+            case DATE :
+                return valueFactories.getDateFactory().create(value);
+            case BOOLEAN :
+                return valueFactories.getBooleanFactory().create(value);
+            case URI :
+                return valueFactories.getUriFactory().create(value);
+            case REFERENCE :
+                return valueFactories.getReferenceFactory().create(value);
+            case SIMPLEREFERENCE :
+                return valueFactories.getSimpleReferenceFactory().create(value);
+            case WEAKREFERENCE :
+                return valueFactories.getWeakReferenceFactory().create(value);
+            default :
+                return value;
+        }
+    }
+    
+    /**
+     * Converts array of values.
+     */
+    protected Object[] cast(Object[] values) {
+        Object[] res = new Object[values.length];
+        for (int i = 0; i < res.length; i++) {
+            res[i] = cast(values[i]);
+        }
+        return res;
+    }
+
+    /**
+     * Converts array of values.
+     */
+    protected Object[] cast(Collection values) {
+        Object[] res = new Object[values.size()];
+        int i = 0;
+        for (Object value : values) {
+            res[i++] = cast(value);
+        }
+        return res;
+    }
+    
+}
--- a/index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndexColumns.java
+++ b/index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndexColumns.java
@@ -1,0 +1,167 @@
+/*
+ * ModeShape (http://www.modeshape.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.modeshape.jcr.index.elasticsearch;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import org.modeshape.jcr.ExecutionContext;
+import org.modeshape.jcr.api.index.IndexColumnDefinition;
+import org.modeshape.jcr.api.index.IndexDefinition;
+import org.modeshape.jcr.index.elasticsearch.client.EsRequest;
+import org.modeshape.jcr.value.PropertyType;
+import static org.modeshape.jcr.value.PropertyType.BINARY;
+import static org.modeshape.jcr.value.PropertyType.BOOLEAN;
+import static org.modeshape.jcr.value.PropertyType.DATE;
+import static org.modeshape.jcr.value.PropertyType.DECIMAL;
+import static org.modeshape.jcr.value.PropertyType.DOUBLE;
+import static org.modeshape.jcr.value.PropertyType.LONG;
+
+/**
+ * Set of index columns where each column provides column's type 
+ * specific conversation.
+ * 
+ * 
+ * @author kulikov
+ */
+public class EsIndexColumns {
+    //list of columns
+    private final Map<String, EsIndexColumn> columns = new HashMap<String, EsIndexColumn>();
+    
+    /**
+     * Creates new set.
+     * 
+     * @param cols set of columns.
+     */
+    public EsIndexColumns(EsIndexColumn... cols) {
+        for (int i = 0; i < cols.length; i++) {
+            columns.put(cols[i].getName(), cols[i]);
+        }
+    }
+    
+    /**
+     * Creates new set of columns using index definition.
+     * 
+     * @param context execution context 
+     * @param defn index definition.
+     */
+    public EsIndexColumns(ExecutionContext context, IndexDefinition defn) {
+        EsIndexColumn[] cols = new EsIndexColumn[defn.size()];
+        for (int i = 0; i < cols.length; i++) {
+            IndexColumnDefinition idef = defn.getColumnDefinition(i);
+            columns.put(idef.getPropertyName(),
+                    new EsIndexColumn(context, idef.getPropertyName(), idef.getColumnType()));
+        }
+    }
+    
+    /**
+     * Gets column with given name.
+     * 
+     * @param name the name of column in index or name of pseudo column like
+     * lower case, upper case or length.
+     * 
+     * @return real column definition related to the specified name
+     */
+    public EsIndexColumn column(String name) {
+        return columns.get(noprefix(name, 
+                EsIndexColumn.LENGTH_PREFIX,
+                EsIndexColumn.LOWERCASE_PREFIX,
+                EsIndexColumn.UPPERCASE_PREFIX));
+    }
+    
+    /**
+     * Cuts name prefix for the pseudo column case.
+     * 
+     * @param name the name of the column or pseudo column.
+     * @param prefix the list of prefixes to cut.
+     * @return the name without prefix.
+     */
+    private String noprefix(String name, String... prefix) {
+        for (int i = 0; i < prefix.length; i++) {
+            if (name.startsWith(prefix[i])) {
+                name = name.replaceAll(prefix[i], "");
+            }
+        }
+        return name;
+    }
+    
+    /**
+     * List of real columns.
+     * 
+     * @return collection of the columns.
+     */
+    public Collection<EsIndexColumn> columns() {
+        return columns.values();
+    }
+
+    /**
+     * Provides Elasticsearch mapping definition for the given type and 
+     * this columns.
+     * 
+     * @param type the type name for wich this mapping will be applied.
+     * @return mappings definition in the JSON format.
+     */
+    public EsRequest mappings( String type ) {
+        EsRequest mappings = new EsRequest();
+            EsRequest mappingsValue = new EsRequest();
+                EsRequest mtype = new EsRequest();
+                    EsRequest properties = new EsRequest();       
+                    for (EsIndexColumn col : columns()) {
+                        properties.put(col.getName(), fieldMapping(col.getType()));
+                        properties.put(col.getLowerCaseFieldName(), fieldMapping(PropertyType.STRING));
+                        properties.put(col.getUpperCaseFieldName(), fieldMapping(PropertyType.STRING));
+                        properties.put(col.getLengthFieldName(), fieldMapping(PropertyType.LONG));
+                    }        
+                mtype.put("properties", properties);
+            mappingsValue.put(type, mtype);
+        mappings.put("mappings", mappingsValue);
+        
+        return mappings;
+    }
+    
+    /**
+     * Creates mapping definition for the specified column type.
+     * 
+     * @param type
+     * @return 
+     */
+    private EsRequest fieldMapping( PropertyType type ) {
+        EsRequest mappings = new EsRequest();
+        switch (type) {
+            case BINARY:
+                mappings.put("type", "binary");
+                break;
+            case BOOLEAN:
+                mappings.put("type", "boolean");
+                break;
+            case DATE:
+                mappings.put("type", "long");
+                break;
+            case LONG:
+            case DECIMAL:
+                mappings.put("type", "long");
+                break;
+            case DOUBLE:
+                mappings.put("type", "double");
+                break;
+            default:
+                mappings.put("type", "string");
+                mappings.put("analyzer", "whitespace");
+        }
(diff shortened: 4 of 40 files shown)


index-providers/modeshape-elasticsearch-index-provider/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+    <!-- The groupId and version values are inherited from parent -->
+    <artifactId>modeshape-elasticsearch-index-provider</artifactId>
+    <packaging>jar</packaging>
+    <name>ModeShape Elasticsearch Index Provider</name>
+    <description>ModeShape index provider that uses Elasticsearch engine</description>
+    <url>http://www.modeshape.org</url>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.elasticsearch</groupId>
```
hchiorean: this should be defined in the parent POM

index-providers/modeshape-elasticsearch-index-provider/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+            <artifactId>elasticsearch</artifactId>
+            <version>1.7.3</version>
+        </dependency>            
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-core</artifactId>
+            <version>4.10.4</version>
```
hchiorean: this should be extracted into a property, otherwise it's duplicated across all dependencies

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EmptyResults.java
```diff
@@ -0,0 +1,42 @@
...
+package org.modeshape.jcr.index.elasticsearch;
+
+import org.modeshape.jcr.spi.index.Index;
+import org.modeshape.jcr.spi.index.ResultWriter;
+
+/**
+ *
+ * @author kulikov
+ */
+public class EmptyResults implements Index.Results {
```
hchiorean: This class it not needed. Instead, [this](https://github.com/okulikov/modeshape/blob/MODE-2162-b/modeshape-jcr/src/main/java/org/modeshape/jcr/index/local/Operations.java#L57) should be "refactored out" and placed as a constant (`public static final`) in the `Results` interface from the SPI.

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndexColumns.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Map;
+
+/**
+ * Set of index columns.
+ * 
+ * @author kulikov
+ */
+public class EsIndexColumns {
+    
+    private Map<String, EsIndexColumn> columns = new HashMap();
```
hchiorean: can this be `final` ?

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndexProvider.java
```diff
@@ -0,0 +1,197 @@
...
+            if (client != null) {
+                client.close();
+            }
+
+            //stop node
+            if (esNode != null) {
+                esNode.stop();
+            }
+        } catch (Exception e) {
+        }
```
hchiorean: shouldn't the exception be logged at least ?

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndexProvider.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @Override
+    protected void postShutdown() {
+        logger().debug("Shutting down the elasticsearch index provider '{0}' in repository '{1}'", getName(), getRepositoryName());
+        try {
+            client.close();
+            esNode.stop();
+        } catch (Exception e) {
+        }
```
hchiorean: shouldn't the exception be logged ?

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndexProvider.java
```diff
@@ -0,0 +1,197 @@
...
+import org.modeshape.jcr.spi.index.IndexCostCalculator;
+import org.modeshape.jcr.spi.index.provider.IndexProvider;
+import org.modeshape.jcr.spi.index.provider.IndexUsage;
+import org.modeshape.jcr.spi.index.provider.ManagedIndexBuilder;
+
+/**
+ *
+ * @author kulikov
+ */
+public class EsIndexProvider extends IndexProvider {
```
hchiorean: this class should also override the ` protected int getCostEstimate()` method from the base class, making sure it returns a `REMOTE` cost. In other words, if there's a local index (MapDB or Lucene) that covers the same property, it should be preferred over the `REMOTE` one.

hchiorean:
In addition to the unit tests specific to the ES provider, there should be an additional test case which extends `LocalIndexProviderTest`. This test case covers a lot of cases which can flush out potential problems with the provider.

This test case _may_ contain additional tests specific to ES: for example testing that when both a local and an ES index is defined over a property, the local index is preferred because of the lower cost. 

The [Lucene index provider](https://github.com/okulikov/modeshape/blob/MODE-2162-b/index-providers/modeshape-lucene-index-provider/src/test/java/org/modeshape/jcr/index/lucene/LuceneIndexProviderTest.java) does a similar thing.

hchiorean:
@okulikov: after sorting all the comments (including @rhauch's if he has any), can you please create a separate PR for `4.x` ? 
`master` and `4.x` have already starting diverging significantly and this feature has to be present in both branches. thanks

hchiorean:
another question: this PR does not cover any work required for the WF kit integration or the general `modeshape-distribution` integration. Will this be a separte PR ?

okulikov:
@hchiorean , `EsIndexProviderTest` extends `LocalIndexProviderTest`

hchiorean:
right, sorry I missed it.

rhauch:
> another question: this PR does not cover any work required for the WF kit integration or the general `modeshape-distribution` integration. Will this be a separte PR ?

@okulikov, any reply? Also would be great to see at least an acknowledgement on @hchiorean's other line notes, or, if you don't agree with his comment, a line note that explains why.

okulikov:
@hchiorean , @rhauch Fixed notes. Let's do WF kit in this PR as well. I just have some issues with testing.

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndex.java
```diff
@@ -0,0 +1,327 @@
...
+
+    @Override
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public Results filter(IndexConstraints constraints) {
+        FilterBuilder filter = operations.createFilter(constraints.getConstraints(), constraints.getVariables());
+//        try {
```
hchiorean: shouldn't these comments be removed ?

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndexProvider.java
```diff
@@ -0,0 +1,204 @@
...
+    protected int getCostEstimate() {
+        return IndexCostCalculator.Costs.REMOTE;
+    }    
+    
+    @Override
+    public void validateProposedIndex(ExecutionContext context,
+            IndexDefinition defn,
+            NodeTypes.Supplier nodeTypeSupplier,
+            Problems problems) {
+        // first perform some custom validations
```
hchiorean: should this be overridden ? In other words, is it intentional for the ES index provider to not perform any validations ?

okulikov: Lets' keep it as place where to do such validations in the future if it will needed.

okulikov:
WF kit and binary distribution part already included into PR

hchiorean:
@okulikov: the changes look fine, with some additional comments (see above). 

Also, a couple of more general questions:
1. why isn't the ElasticSearch version the latest one - `2.1.0` (as opposed to `1.7.3`) ? This latest version also uses Lucene 5.3.1, which is exactly what the Lucene index provider uses
2. an additional integration test should be added which validates that the ES provider works in WF. [This](https://github.com/ModeShape/modeshape/blob/master/integration/modeshape-jbossas-integration-tests/src/main/resources/kit/jboss-wf9/standalone/configuration/standalone-modeshape.xml#L330) is where the new index provider should be added and tested. This will most likely require the same test dependencies as the `es-index-provider` module, so they should be added to the `modeshape-jbossas-integration-tests` module and the Arquillian code from the [unit test class](https://github.com/ModeShape/modeshape/blob/master/integration/modeshape-jbossas-integration-tests/src/test/java/org/modeshape/test/integration/QueryIntegrationTest.java#L59) updated.
3. Please rebase the code on the latest `master` or `4.x` because it cannot be automatically merged.

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndexProvider.java
hchiorean: ok; I forgot the base class only provides an empty implementation since the default validations are performed always.

okulikov:
@hchiorean answering in the same order:
1. The reason is that older version API provides `Filters` in addition to `Queries` and without using `Filters` I didn't find any way to pass 'all' tests inherited from 'LocalIndexProviderTest`. It very often return ether less results the test case is expecting or more. Let's for now stick to this version and later I can try to reach latest version;
2. The integration test will require access to the ES cluster. If I am right the dependencies should be added to the 'org.elasticsearch' WF module definition. Is it possible to perform it for test purposes only?
3. Will do.

hchiorean:
>1. The reason is that older version API provides Filters in addition to Queries and without using Filters I didn't find any way to pass 'all' tests inherited from 'LocalIndexProviderTest`. It very often return ether less results the test case is expecting or more. Let's for now stick to this version and later I can try to reach latest version;

That may be so, but I'm not sure we want to support an older version of Elastic Search, for numerous reasons. Not only there have been a number of bug fixes, but Lucene 5.x is superior to Lucene 4.10
@rhauch thoughts ?

>2. The integration test will require access to the ES cluster. If I am right the dependencies should be added to the 'org.elasticsearch' WF module definition. Is it possible to perform it for test purposes only?

If the test only works with a real cluster, then no, we should definitively not be adding these dependencies to the WF module. I thought that the unit testing done as part of ES index provider artifact uses some sort of in-memory/embedded mode (similar to Cassandra testing for example). If that's not the case, then there's not much more we can do.

okulikov:
@hchiorean , @rhauch I can try a bit more to fix problems with latest version. Any thoughts?

hchiorean:
my vote is on using the latest Elastic Search version `2.1.0` since this version/API will be evolved in the future by the ES team.

rhauch:
@hchiorean, @okulikov: I think we should use the latest version of ES.

> > The integration test will require access to the ES cluster. If I am right the dependencies should be added to the 'org.elasticsearch' WF module definition. Is it possible to perform it for test purposes only?

> If the test only works with a real cluster, then no, we should definitively not be adding these dependencies to the WF module. I thought that the unit testing done as part of ES index provider artifact uses some sort of in-memory/embedded mode (similar to Cassandra testing for example). If that's not the case, then there's not much more we can do.

Actually, if the provider can only be tested with an in-memory ES server, then we should at least do that in this PR. However, if really the only way to test this is with a real ES cluster, we can use Maven to start a Docker container with ES, have the tests use it, and then have Maven shut down the container. Might be better to do that in a separate PR, though. We'll likely want to use Docker in other modules as well -- for example, to test with PostgreSQL, MySQL, or other DBMSes.

okulikov:
@hchiorean , @rhauch So I will focus on latest ES version first with PR and then prepare another PR with integration test

rhauch:
@okulikov, Sorry if I wasn't clear. I meant that if you can do any integration test with ES without Docker (either because ES can be run in-process or started/managed by the test itself), then that should be done as part of this PR. After all, that's the only way to verify that this feature is working correctly. OTOH, if the only real way to test integration with ES is via Docker, then IMO that can be done in a subsequent PR.

Perhaps you've already seen thins: it appears that the tests can start and manage the ES cluster; see https://www.elastic.co/guide/en/elasticsearch/reference/current/integration-tests.html. If that is the case, that should be the preferred approach, since it will be the simplest (and probably fastest).

okulikov:
@rhauch , we are on the same page. I hope that moving to latest version will automatically help to run intergration-test without Docker since all necessary Lucene dependencies  already deployed as WF module.

hchiorean:
> OTOH, if the only real way to test integration with ES is via Docker, then IMO that can be done in a subsequent PR.

not sure how easy it would be to integrate Docker with Maven and tie it up together with the build system (while also keeping a full `clean install -Passembly` run in reasonable time). As an alternative, I think we can explore the Docker integration for either the `examples` or `quickstarts` repos

okulikov:
One test case remains

okulikov:
@hchiorean , @rhauch can we skip one test case `shouldUseIndexesAfterRestarting` ? ES cluster switches to YELLOW state during repository restart.  It was not happening with previous version.

hchiorean:
if you mean just for the `ESIndexProviderTest`case, that's fine IMO (i.e. you override this and ignore it just in this particular test case)

okulikov:
yes, exactly

index-providers/modeshape-elasticsearch-index-provider/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+        </dependency> 
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-queryparser</artifactId>
+            <version>${version.elasticsearch.test}</version>
+            <scope>test</scope>
+        </dependency>         <!-- 
+            Required to text FTS on binary properties...
+        -->
+        <!-- dependency>
```
hchiorean: if this is not used, it should be removed

okulikov:
@rhauch , @hchiorean Done without Docker usage.

deploy/jbossas/kit/jboss-wf9/org/elasticsearch/2.1.0/module.xml
```diff
@@ -0,0 +1,51 @@
...
+	<resource-root path="jackson-dataformat-smile-2.6.2.jar" />
+    </resources>
+    <dependencies>
+        <module name="javax.api" export="true"/>
+        <module name="sun.jdk" export="true" >
+           <imports>
+               <include path="sun/misc/Unsafe" />
+           </imports>
+        </module>
+	<module name="org.apache.lucene" slot="5.3.1"/>
```
hchiorean: these should use the version properties defined in the parent pom, it makes it easier to maintain that way

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/TransportAddress.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import org.elasticsearch.common.transport.InetSocketTransportAddress;
+
+/**
+ *
+ * @author kulikov
+ */
+public class TransportAddress {
```
hchiorean: what's the purpose of this class ? Hardcoding a port doesn't seem right...

index-providers/modeshape-elasticsearch-index-provider/src/test/java/org/modeshape/jcr/index/elasticsearch/EsIndexProviderTest.java
```diff
@@ -0,0 +1,127 @@
...
+import org.junit.Test;
+import org.modeshape.common.util.FileUtil;
+import org.modeshape.jcr.EsValidateQuery;
+import org.modeshape.jcr.LocalIndexProviderTest;
+import org.modeshape.jcr.ValidateQuery;
+import org.modeshape.jcr.api.JcrTools;
+import org.modeshape.jcr.api.query.Query;
+
+/**
+ * Unit test for {@link LuceneIndexProvider}. Since this is a local provider in term of repository locality, we want to run
```
hchiorean: This JavaDoc should be fixed, it's not accurate.

integration/modeshape-jbossas-integration-tests/src/main/webapp/WEB-INF/jboss-deployment-structure.xml
```diff
@@ -0,0 +1,10 @@
+<jboss-deployment-structure>
+    <deployment>
+	<dependencies>
+	    <module name="javax.jcr"/>
+	    <module name="org.elasticsearch" slot="2.1.0" services="import" />
```
hchiorean: this should use the Maven version property.

integration/modeshape-jbossas-integration-tests/src/main/webapp/WEB-INF/jboss-deployment-structure.xml
```diff
@@ -0,0 +1,10 @@
+<jboss-deployment-structure>
```
hchiorean: why is this explicit dependency file required by the test ? 
The WF config specifies a module dependency in the actual index-provider element:
```
<index-provider name="elasticsearch" classname="org.modeshape.jcr.index.elasticsearch.EsIndexProvider" module="org.modeshape.index-provider.elasticsearch"/>

as such, the CP should be resolved at runtime correctly.

Comment (okulikov) [inline integration/modeshape-jbossas-integration-tests/src/main/webapp/WEB-INF/jboss-deployment-structure.xml]: @hchiorean , This explicit dependency allows to start local ES node before running test

File: index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/TransportAddress.java (hunk 12 above)
Comment (okulikov) [inline index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/TransportAddress.java]: This class is used for parsing addresses and ports of the ES cluster nodes (actually it is separated as dedicated class because it is easy to test its feature rather then implement/test as private method inside `EsIndexProvider` class). The default value is localhost:9300. No hardcoded  ports.

Comment (hchiorean) [inline index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/TransportAddress.java]: u'r right, I didn't notice that constant is just the default value
still, I'm not 100% sure a separate class is required for testing essentially a `string.split` method

Comment (okulikov) [inline index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/TransportAddress.java]: @hchiorean , Just to be sure. At this moment I can do a copy/paste of the method and discard the class.

**Comment by hchiorean** [comment]
@okulikov this looks good, the only thing that's left is using the `${}` version properties in the `module.xml` file
Also, can you please squash the 3 commits into 1

@rhauch are you ok with the PR ?

File: deploy/jbossas/kit/jboss-wf9/org/apache/lucene/5.3.1/module.xml [hunk 16]
```diff
@@ -23,8 +23,16 @@
         <resource-root path="lucene-queries-${version.org.apache.lucene}.jar" />
         <resource-root path="lucene-queryparser-${version.org.apache.lucene}.jar" />
         <resource-root path="lucene-sandbox-${version.org.apache.lucene}.jar" />
+	<resource-root path="lucene-join-${version.org.apache.lucene}.jar" />
+	<resource-root path="lucene-highlighter-${version.org.apache.lucene}.jar" />
+	<resource-root path="lucene-memory-${version.org.apache.lucene}.jar" />
+	<resource-root path="lucene-suggest-${version.org.apache.lucene}.jar" />
+	<resource-root path="lucene-misc-${version.org.apache.lucene}.jar" />
+	<resource-root path="lucene-grouping-${version.org.apache.lucene}.jar" />
+	<resource-root path="lucene-spatial-${version.org.apache.lucene}.jar" />
```
rhauch: Are all of these used in the index provider? Do they all come with the standard Lucene install, or were some added from other downloads? For example, I'm surprised if we use spatial or highlighter in our index provider.

rhauch: Ah, never mind. I guess the Elastic Search client library probably requires them all, even if we don't directly use those aspects of Lucene.

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndexColumn.java
```diff
@@ -0,0 +1,254 @@
...
+            case PATH:
+            case NAME:
+            case STRING:
+            case REFERENCE:
+            case SIMPLEREFERENCE:
+            case WEAKREFERENCE:
+            case URI:
+                return valueFactories.getStringFactory().create(value);
+            case DATE :
+                return ((DateTime) value).getMilliseconds();
```
rhauch: There is no `default` case here. What about longs or binaries or other `PropertyType`s?

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndexColumn.java
```diff
@@ -0,0 +1,254 @@
...
+                return valueFactories.getBooleanFactory().create(value);
+            case URI :
+                return valueFactories.getUriFactory().create(value);
+            case REFERENCE :
+                return valueFactories.getReferenceFactory().create(value);
+            case SIMPLEREFERENCE :
+                return valueFactories.getSimpleReferenceFactory().create(value);
+            case WEAKREFERENCE :
+                return valueFactories.getWeakReferenceFactory().create(value);
+        }
```
rhauch: Again, this switch is incomplete. IIUC, this should even compile with the checkstyle rules.

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/Operations.java
```diff
@@ -0,0 +1,351 @@
...
+                    
+                    return builder;
+            }
+        }
+
+        if (constraint instanceof SetCriteria) {
+            SetCriteria setCriteria = (SetCriteria)constraint;
+            
+            String field = (String) operand(setCriteria.getOperand()).apply(setCriteria.getOperand(), variables);
+            ArrayList list = new ArrayList();
```
rhauch: We should never used untyped generics.

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/Operations.java
```diff
@@ -0,0 +1,351 @@
...
+            for (StaticOperand so : setCriteria.getValues()) {
+                Object vals = operand(so).apply(field, so, variables);
+                if (vals instanceof Object[]) {
+                    list.addAll(Arrays.asList((Object[]) vals));
```
rhauch: The values will be of the type used within the query system. Are you sure these values don't need to be converted before they're passed into the query?

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/Operations.java
```diff
@@ -0,0 +1,351 @@
...
+            
+            if (value instanceof Collection) {
+                return col.cast((Collection) value);
+            }
+            
+            return col.cast(variables.get(op.getBindVariableName()));
+        }
+    };
+
+    private final OperandBuilder LOWER_CASE_BUILDER = new OperandBuilder<LowerCase>() {
```
rhauch: If these are immutable constants, they should be marked as static. If not, their name shouldn't be uppercased.

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/SearchResults.java
```diff
@@ -0,0 +1,92 @@
...
+    @Override
+    public boolean getNextBatch(ResultWriter writer, int batchSize) {
+        try {
+            response = searchBuilder
+                    .setSize(batchSize)
+                    .setFrom(pos)
+                    .execute()
+                    .actionGet();
+        } catch (Exception e) {
+            e.printStackTrace();
```
rhauch: Is there no better way to handle this exception? At least it should be logged and not printed.

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/SearchResults.java
```diff
@@ -0,0 +1,92 @@
...
+     * @return total hits matching search request.
+     */
+    public long getCardinality() {
+        try {
+            response = searchBuilder
+                    .execute()
+                    .actionGet();
+            return response.getHits().getTotalHits();
+        } catch (Exception e) {
+            return 0;
```
rhauch: So if there is a network error, this method will return `0`? Seems strange.

index-providers/modeshape-elasticsearch-index-provider/src/test/java/org/modeshape/jcr/index/elasticsearch/EsIndexProviderTest.java
```diff
@@ -0,0 +1,126 @@
...
+                .rowCount(3L)
+                .useIndex("longValues")
+                .hasNodesAtPaths("/2", "/4", "/1")
+                .validate(query, query.execute());
+    }
+    
+    @Test
+    @Override
+    public void shouldUseIndexesAfterRestarting() throws Exception {
+    }
```
rhauch: Can this test not be implemented or at least specialized so that we know that the index provider after restart will connect properly to a still-running Elasticsearch node?

rhauch:
@okulikov, @hchiorean: see my questions above.

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndexColumn.java
okulikov: No conversation will happen. The last statement returns value as is

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndexColumn.java
okulikov: Same as above. If type is outside of the switch scope the same value will be returned.

index-providers/modeshape-elasticsearch-index-provider/src/test/java/org/modeshape/jcr/index/elasticsearch/EsIndexProviderTest.java
okulikov: Local ES node used for testing switches to "YELLOW" state during repository restart (I guess this is ES bug )so we don't have still-running ES node during this test.

rhauch: Then perhaps add a comment stating such?

okulikov: makes sense

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndexColumn.java
rhauch: Not having the switch include all possible values is very much preferred, and using the `default` case is a secondary option. However, doing neither is IMO bad form since it's implicitly providing a default case and not using Java's built-in `default` case. Yes, you'll need to throw an exception as the last line of the method to signal that getting there should never happen; even the ugliness of throwing an exception is better than an incomplete switch.

okulikov:
@rhauch , @hchiorean Fixed commented lines

hchiorean:
@okulikov: I'm running the integrated PR code on my machine and there are some things which have to be fixed:

1. https://gist.github.com/hchiorean/c501770624ee74775438 - a lot of javac warnings due unchecked/raw calls. If any/some of those calls are due to the ES API, then they should be avoided with `SupressWarnings`. Otherwise generic types should be used.

2. the time required to run the `EsIndexProviderTest` on my machine is extremely high: `7 minutes`. This is something which has to be fixed, because IMO it's simply not feasible to add a module in the MS general build which increases the build duration by this much.
For example, in the `EsIndexProviderTest` class, why is an `org.elasticsearch.node.Node` started and stopped with each test ? This doesn't seem right: the ES cluster should be started once before all the tests and stopped after all the tests (i.e. `BeforeClass` and `AfterClass`). This would also eliminate the need to `Thread.sleep(500)` between each test, just to wait for a shutdown

3. the integration tests are failing: https://gist.github.com/hchiorean/33cdc73c23ebbe5265dc meaning that the WF integration is not correct

4. ES client - this is something which hasn't been discussed so far, but it should be: 
the only thing ModeShape should require for this integration is the ES (or another) API that *communicates with the cluster* and nothing else. In other words, *no dependency from the following graph should be present in the ModeShape ES distributions*: 
```
 +- org.apache.lucene:lucene-core:jar:5.3.1:compile
 +- org.apache.lucene:lucene-backward-codecs:jar:5.3.1:compile
 +- org.apache.lucene:lucene-analyzers-common:jar:5.3.1:compile
 +- org.apache.lucene:lucene-queries:jar:5.3.1:compile
 +- org.apache.lucene:lucene-memory:jar:5.3.1:compile
 +- org.apache.lucene:lucene-highlighter:jar:5.3.1:compile
 +- org.apache.lucene:lucene-queryparser:jar:5.3.1:compile
 |  \- org.apache.lucene:lucene-sandbox:jar:5.3.1:compile
 +- org.apache.lucene:lucene-suggest:jar:5.3.1:compile
 |  \- org.apache.lucene:lucene-misc:jar:5.3.1:compile
 +- org.apache.lucene:lucene-join:jar:5.3.1:compile
 |  \- org.apache.lucene:lucene-grouping:jar:5.3.1:compile
 +- org.apache.lucene:lucene-spatial:jar:5.3.1:compile
 |  +- org.apache.lucene:lucene-spatial3d:jar:5.3.1:compile
 |  \- com.spatial4j:spatial4j:jar:0.4.1:compile
 +- com.google.guava:guava:jar:18.0:compile
 +- com.carrotsearch:hppc:jar:0.7.1:compile
 +- joda-time:joda-time:jar:2.8.2:compile
 +- org.joda:joda-convert:jar:1.2:compile
 +- com.fasterxml.jackson.core:jackson-core:jar:2.6.2:compile
 +- com.fasterxml.jackson.dataformat:jackson-dataformat-smile:jar:2.6.2:compile
 +- com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:jar:2.6.2:compile
 |  \- org.yaml:snakeyaml:jar:1.15:compile
 +- com.fasterxml.jackson.dataformat:jackson-dataformat-cbor:jar:2.6.2:compile
 +- io.netty:netty:jar:3.10.5.Final:compile
 +- com.ning:compress-lzf:jar:1.0.2:compile
 +- com.tdunning:t-digest:jar:3.0:compile
 +- org.hdrhistogram:HdrHistogram:jar:2.1.6:compile
 +- commons-cli:commons-cli:jar:1.3.1:compile
 \- com.twitter:jsr166e:jar:1.1.0:compile
```
@rhauch touches on this topic in his previous comment

The client should only send/receive requests and therefore not require any of the above (of course we might need the full lot for testing, but otherwise none should be required). Now, I'm also aware (see https://www.elastic.co/blog/found-java-clients-for-elasticsearch) that ES doesn't provide a dedicated client jar OOTB (which truly sucks IMO). If that means we need to bring in all of the above dependencies, then we need another client. The previous link lists some possible Java alternatives.

One of the main reasons why having this lot of dependencies is not an option, is that this means the ES index provider *will never be productized* since many of those dependencies will most likely never be compatible with the ones from EAP 7.

In conclusion, the MS ES index provider should use a client API which is as lightweight as possible and which doesn't bring in a lot of transitive dependencies.

okulikov:
@hchiorean , I also extremely dislike such long list of dependencies not transparent for me as well as dependencies from the `Lucene` and 'Elasticsearch' jars. And if we will agree I would prefer to implement a simple and lightweight client for our purposes

hchiorean:
@okulikov +1 

Ideally we wouldn't implement everything from scratch. Perhaps we can use some code from/fork https://github.com/searchbox-io/Jest. Looking at it the only dependency we don't need from their codebase is `guava` and perhaps `gson`. Apart from those 2 it's lightweight enough and ALS2 licensed.

@rhauch any thoughts on this topic ? (point 4 from my previous comment)

rhauch:
@hchiorean, @okulikov: is it possible to use the official client and prune nearly all of the dependencies? If we're not using much of that code, then might this work?

okulikov:
@rhauch, @hchiorean  it is not possible. we already are using minimal code and it still depend from all that jars. Also I am thinking that we should not shy to have our own lightweight client. What I understood is that the most "complicated" part is reading and parsing response but in our case it is simple because we need to extract only identifier of the node. Making request suppose only string formating.

rhauch:
@okulikov, I'm fine with trying out our own client, since the requirements are quite narrowly focused and constrained.

okulikov:
@hchiorean , @rhauch, Latest commit contains lightweight ES client which does not depend from any external jars now and works fast. Dependency from `elasticsearch` jar remains only in test scope and it is required to start ES cluster only for unit-tests. 
However, for integration test the problem still remain. What I do is
- create arquillian test case;
- create war deployment using ShrinkWrap where WEB-INF/lib folder contains `elasticsearch-2.1.0` jar with transitive dependencies  `lucene-*-5.3.1.jar`;
- `@before` test-case ES node started, `@after` test-case stopped.

`War` itself looks good. All necessary classes are really in the `WEB-INF/lib` folder.  However ES node doesn't start. Exception indicates that there is a version mess. Reason is that WF provides its own `lucene-4.10` unit with older classes and that classes are on the same class path with web application. I tried all tricks with `manifest.mf`, `jboss-deployment-descriptor`, repackaging... The only working way is to define custom module, put all jars to that module and explicit specify dependency. But this is weird:
- distribution contains custom module which is used only for integration test;
- this custom module contains the same long list of jars.

Another options:
- using [ES testing framework](https://www.elastic.co/guide/en/elasticsearch/reference/2.x/testing-framework.html); It is not available in the current release (only on snapshots);
- using `Docker`; This way looks overcomplicated for running one simple IT test case.

Taking into account that test-case itself enough simple (index to nodes and search with simple query) my suggestion is to implement also our own simple backing node. All the ES specific functionality covered by functional unit-test and the purpose of IT test is that provider is plugged and communication exists. What do you think guys?

hchiorean:
@okulikov: did you try explicitly excluding the older Lucene version ? The version mess is caused by ISPN 7.x bring in its own version of Lucene & Co. Fortunately, these are installed in separate AS modules, so one option is to tweak the `jboss-deployment-descriptor` of the ShrinkWrap WAR and explicitly remove those modules.

see https://developer.jboss.org/message/850892#850892 for an example of dependency descriptor which excludes certain modules

okulikov:
@hchiorean , let me try

okulikov:
@hchiorean, Declarative exclusions did not help. I also tried to pick up dependencies from `ShrikWrap` and load required classes with custom class loader and also without any luck. 
In latest suggested  another way - prepare and deploy elasticsearch WF module using maven especially for integration tests only. Outside of the integration test this module does not exists.

index-providers/modeshape-elasticsearch-index-provider/src/test/java/org/modeshape/jcr/index/elasticsearch/EsIndexProviderTest.java
```diff
@@ -0,0 +1,132 @@
...
+        startRepositoryWithConfiguration(repositoryConfiguration());
+        printMessage("Started repository...");
+        tools = new JcrTools();
+    }
+
+    @Override
+    public void afterEach() throws Exception {
+        try {
+            stopRepository();
+        } catch (Exception e) {
```
hchiorean: should this exception be left to bubble up and fail the test ? If `afterEach` fails it means the repository was not shutdown properly

index-providers/modeshape-elasticsearch-index-provider/src/test/java/org/modeshape/jcr/index/elasticsearch/EsIndexTest.java
```diff
@@ -0,0 +1,369 @@
...
+                .node();
+        client = new EsClient("localhost", 9200);
+        index = new EsIndex(client, cols, context, "test", "workspace");
+
+        initialIndex();
+        
+        try {
+            Thread.currentThread().sleep(1000);
+        } catch (Exception e) {
+        }
```
hchiorean: the only possible checked exception here is `InterruptedException` which if thrown, should be handled accordingly (i.e. interrupted status cleared and perhaps rethrown)

index-providers/modeshape-elasticsearch-index-provider/src/test/java/org/modeshape/jcr/index/elasticsearch/EsIndexTest.java
```diff
@@ -0,0 +1,369 @@
...
+
+    @AfterClass
+    public static void tearDownClass() {
+        index.shutdown(true);
+        esNode.close();
+        FileUtil.delete("target/data");
+    }
+
+    @Before
+    public void setUp() {
```
hchiorean: why have  this method at all if it's empty ?

index-providers/modeshape-elasticsearch-index-provider/src/test/java/org/modeshape/jcr/index/elasticsearch/EsIndexTest.java
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    @After
+    public void tearDown() {
```
hchiorean: why have  this method at all if it's empty ?

index-providers/modeshape-elasticsearch-index-provider/src/test/java/org/modeshape/jcr/index/elasticsearch/EsIndexTest.java
```diff
@@ -0,0 +1,369 @@
...
+        index.add(key("key6"), def1.getName(), "the quick Brown fox jumps over to the dog in at the gate");
+        index.add(key("key6"), def2.getName(), 55);
+        index.add(key("key6"), def3.getName(), new String[]{"m", "n", "o"});
+
+        index.add(key("key7"), def5.getName(), "asd-sdf-dfg");
+        
+        try {
+            Thread.currentThread().sleep(1000);
+        } catch (Exception e) {
+        }
```
hchiorean: again, interruption should not be ignored

integration/modeshape-jbossas-integration-tests/pom.xml
```diff
@@ -231,6 +241,20 @@
...
                         </configuration>
                     </execution>
+                    
+                    <execution>
+                        <id>copy-dependencies</id>
+                        <phase>process-test-classes</phase>
+                        <goals>
+                            <goal>copy-dependencies</goal>
+                        </goals>                        
+                        <configuration>
```
hchiorean: shouldn't this be configured to only copy the `org.elasticsearch` artifact-related dependencies ? In other words, don't copy all the module dependencies. Also, the generic `libs` name can then be changed to smth like `elasticsearchLibs` or something similar

integration/modeshape-jbossas-integration-tests/pom.xml
```diff
@@ -259,6 +283,45 @@
...
+                            <outputDirectory>${project.build.directory}/${jboss.server.root.folder}/modules/org/es-test/2.1.0</outputDirectory>
+                            <overwrite>true</overwrite>
+                            <resources>
+                                <resource>
+                                    <directory>src/test/resources/modules/es-test</directory>
+                                    <include>module.xml</include>
+                                </resource>
+                                <resource>
+                                    <directory>${project.build.directory}/libs</directory>
```
hchiorean: if the above `copy-dependencies` goal is configured to only include `elasticsearch` related artifacts, this won't need this entire list of specific artifacts to include. Instead, in can include everything from the `elasticsearchLibs` dir.

hchiorean:
@okulikov: this looks ok to me with some minor comments (see above)

rhauch:
@okulikov: I agree with @hchiorean: those smaller issues should be addressed, and then this looks good.

okulikov:
@hchiorean , @rhauch, New commit with fixes for the commented issues.

integration/modeshape-jbossas-integration-tests/pom.xml
okulikov: @hchiorean , in case of artifact-related configuration maven ignores transitive dependencies. Later, for WF module I need to specify jar names so for this particular case more useful to follow rule based on jar names. 

I did change dependency output directory name from generic `lib` to more specific `eslib`

hchiorean:
@okulikov: the ES integration tests are failing on my machine when trying to build. See https://gist.github.com/hchiorean/a9ff8edc7194e1c3a821 for more details

okulikov:
@hchiorean , can you please more lines from above. It is not visible from the gist status of ES cluster.
Exception shows that ES cluster is not active.

hchiorean:
@okulikov a couple of things:

1.  the `Connection refused: connect` exception is raised *not* when running the `QueryIntegrationTest` but `DatasourceIntegrationTest`. See [this gist](https://gist.github.com/hchiorean/a9ff8edc7194e1c3a821) which is the full output from the console. The reason is simple: the ES index is defined as part of the `artifacts` repository which is used from different test cases. However, only the  `QueryIntegrationTest` contains the necessary boiler plating for starting/stopping the ES cluster. As such, the ES index integration test _should use a separate, dedicated repository_, not one which is shared by other test cases

2.  the test failure from `QueryIntegrationTest` may or may not be related to the above. However if you look at [this gist](https://gist.github.com/hchiorean/d796cf1eed1abe4c1b00) (which is again the full console output) you can see that the ES cluster starts up correctly but the test fails with a regular failure.

maybe correcting (1) will fix (2), but the fix need to be tested by running all the tests from the `modeshape-integration-test` module.

okulikov:
@hchiorean , Agree with (1). Let me fix it fist.

okulikov:
@hchiorean , updated commit. Moved Elasticsearch index test to a dedicated test case and repository. Also I added 3 seconds delay because the only visible reason for (2) failure is thread racing. On my side it runs without failure constantly.

hchiorean:
@okulikov: the test also passes/passed on my machine when ran in isolation. However, it failed constantly when running `mvn clean install -Pintegration` on the project root.

integration/modeshape-jbossas-integration-tests/src/test/java/org/modeshape/test/integration/QueryIntegrationTest.java
```diff
@@ -54,14 +54,15 @@
...
+                .workOffline()
+                .loadPomFromFile("pom.xml")
+                .resolve("org.modeshape:modeshape-jcr:test-jar:tests:?").withTransitivity().asFile();
+
         return ShrinkWrap.create(WebArchive.class, "query-test.war").addAsLibraries(testDeps)
-                         .addAsWebInfResource(EmptyAsset.INSTANCE, ArchivePaths.create("beans.xml"))
-                         .addAsResource(new File("src/test/resources/text-extractor/text-file.txt"))
-                         .setManifest(new File("src/main/webapp/META-INF/MANIFEST.MF"));
+                .addAsWebInfResource(EmptyAsset.INSTANCE, ArchivePaths.create("beans.xml"))
+                .addAsWebInfResource(new File("src/main/webapp/WEB-INF/jboss-deployment-structure.xml"), ArchivePaths.create("jboss-deployment-structure.xml"))
```
hchiorean: since the ES test have been moved to another class, this isn't required here, right ?

integration/modeshape-jbossas-integration-tests/src/test/java/org/modeshape/test/integration/EsIndexIntegrationTest.java
```diff
@@ -0,0 +1,147 @@
...
+        Settings localSettings = Settings.settingsBuilder()
+                .put("http.enabled", true)
+                .put("number_of_shards", 1)
+                .put("number_of_replicas", 1)
+                .put("path.home", "target/data")
+                .build();
+
+        //configure Elasticsearch node
+        esNode = NodeBuilder.nodeBuilder().settings(localSettings).local(false).build().start();
+        Thread.currentThread().sleep(3000);
```
hchiorean: I think 1000 (1sec) is enough here since this is the same value that the standalone tests use (from the es-index-provoider module)

hchiorean:
@okulikov: with the latest changes the build runs fine; I just had a couple of minor test comments

okulikov:
@hchiorean , @rhauch , Updated with accordance to latest comments.

hchiorean:
@okulikov the build still passes, so it looks good

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndex.java
```diff
@@ -0,0 +1,308 @@
...
+        CheckArg.isNotNull(values, "values");
+
+        EsIndexColumn column = columns.column(propertyName);
+        assert column != null : "Unexpected column for the index " + name();
+
+        try {
+            EsDocument doc = findOrCreateDoc(nodeKey);
+            putValues(doc, column, values);
+            client.storeDocument(name(), workspace, nodeKey, doc);
+            client.refresh(name());
```
rhauch: @okulikov, can you explain why `client.refresh(...)` is called after every add and remove rather than during the commit of changes to the index? Perhaps I don't understand the API well enough, but ES refresh seems to cause a refresh of the index so that any changes since the last refresh are immediately visible.

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndex.java
```diff
@@ -0,0 +1,308 @@
...
+        doc.put(column.getName(), columnValue);
+        doc.put(column.getLowerCaseFieldName(), lc);
+        doc.put(column.getUpperCaseFieldName(), uc);
+        doc.put(column.getLengthFieldName(), ln);
+    }
+
+    @Override
+    public void commit() {
+        try {
+            client.flush(name());
```
rhauch: Again, perhaps I don't understand the Java API, but IIUC `client.flush(...)` is explicitly telling ES to flush all memory associated with an index to storage, and that ES tries to do this automatically. If we have to do this, then `commit()` is definitely the place to do it. But is it really required? Do we want to force a flush to get the durability of the write to disk?

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/EsIndex.java
rhauch: BTW, I know that `ESClient` is ModeShape-specific, but it seems to be a convenient wrapper around the normal ES API.

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/client/EsDocument.java
```diff
@@ -0,0 +1,133 @@
...
+import org.infinispan.schematic.document.Json;
+import org.infinispan.schematic.internal.document.BasicArray;
+
+/**
+ * Json document used for exchange data with Elasticsearch engine.
+ * 
+ * @author kulikov
+ */
+public class EsDocument {
+    private EditableDocument document = DocumentFactory.newDocument();
```
rhauch: The initialization of `document` should be moved into the default constructor so the other constructor doesn't do it twice. Not sure how much of an impact it will have, but its simple enough to correct.

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/client/EsDocument.java
```diff
@@ -0,0 +1,133 @@
...
+    public Object get(String name) {
+        Object obj = document.get(name);
+        if (obj instanceof BasicArray) {
+           Object[] res = new Object[((BasicArray)obj).size()];
+           for (int i = 0; i < res.length; i++) {
+                res[i] = ((BasicArray)obj).get(i);
+           }
+           return res;
+        }
+        return document.get(name);
```
rhauch: Shouldn't this just return `obj`?

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/client/EsDocument.java
```diff
@@ -0,0 +1,133 @@
...
+import org.infinispan.schematic.document.EditableDocument;
+import org.infinispan.schematic.document.Json;
+import org.infinispan.schematic.internal.document.BasicArray;
+
+/**
+ * Json document used for exchange data with Elasticsearch engine.
+ * 
+ * @author kulikov
+ */
+public class EsDocument {
```
rhauch: I'm curious why this class even exists. It seems like it is less efficient in most read-only cases, since this class wraps an _editable_ document, which may not be needed when the EsDocument is just read.

index-providers/modeshape-elasticsearch-index-provider/src/main/java/org/modeshape/jcr/index/elasticsearch/client/EsDocument.java
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * Gets property value.
+     * 
+     * @param name property name.
+     * @return property value.
+     */
+    public Object get(String name) {
+        Object obj = document.get(name);
+        if (obj instanceof BasicArray) {
+           Object[] res = new Object[((BasicArray)obj).size()];
```
rhauch: Perhaps `Array` (which `BasicArray` implements) also implements `List<Object>`, so it should be possible to just call `array.toArray(new Object[array.size()])`. Perhaps `Array` should provide a more convenient method for doing exactly this.

Although, I still think that if the ES index provider code directly used `Document` and `Array`, this would not really be needed.

okulikov:
@rhauch , PR updated:
- Refresh process moved to index `commit`method; You are right, no need to call each time for add/remove;
- `EsDocument` replaced with two classes dedicated for read and write; 
- `arrays` conversation with more suitable method.

`EsIndex` does all required types conversations using information given by `IndexColumnDefinition` and value factories provided by MS (value factories receive `Object` argument). In case of multi-value property such conversion is applied for each element of an `Object[]` array. Thus there is no need to get/set typed value but It seems to me that it is convenient to introduce wrapper(s) around `Document` and hide a differences between array types (ie `Object[]` and `BasicArray`)

rhauch:
Looks good to me.

hchiorean:
@okulikov: I merged this in `master`, but it does not cherry-pick cleanly in `4.x`.
Can you pls cherry-pick & fix conflicts and then create a separate PR for `4.x` ? Thanks.

PR #1489, opened 2015-11-27
MODE-2537-4.x Fixes the improper cleanup of components in case of an unexpected shutdown

hchiorean:
This also updates some unit tests which failed to appropriately shut down their respective `ExecutorService` instances.


rhauch:
Looks good.

PR #1490, opened 2015-12-02
MODE-2539 Fixes the potential corruption of the system area after a restore operation

hchiorean:
The problem occurs when restoring a 3.x repository (pre indexing) to a 4.x repository (post indexing) and occurs due to the fact that the `/jcr:system/mode:indexes` area has to be initialized only *after* the repository is restarted post-restore. This is because a restore operation can change the `rootNode` key and source key of the running repository.


rhauch:
Looks good.

PR #1491, opened 2015-12-02
MODE-2539-master Fixes the potential corruption of the system area after a restore operation

hchiorean:
Porting of https://github.com/ModeShape/modeshape/pull/1490 to `master`


rhauch:
Looks good.

PR #1492, opened 2015-12-03
MODE-2541 Removes the duplicate closing of the hashing stream used when storing binaries in the FS binary store.

hchiorean:


rhauch:
Looks good.

PR #1494, opened 2015-12-08
MODE-2542 Adds the ability to cluster ModeShape repositories using dedicated JGroups channels

hchiorean:
This a semantical rollback of https://issues.jboss.org/browse/MODE-2079, but with a simplified configuration: ModeShape now has a built-in default for the JGroups configuration file and the WF kit supports both a separate JGroups configuration file and also a built-in server-defined stack.

For the standalone (JSON) configuration there are at maximum 2 parameters:
```javascript
  "clustering" : {
        "clusterName" : "modeshape",
        "configuration" : "config/cluster/jgroups-test-config.xml"
    }
```
or at a minimum:
```javascript
 "clustering" : {}  
```
In the latter case ModeShape will use 2 built-in defaults: `ModeShape-JCR` for the name of the cluster and a TCP-based stack for the JGroups configuration.

For the WF kit, ModeShape will now support 2 configuration styles:
* cluster name and external JGroups file (for users migrating from 4.x)
```xml
<repository name="sample3"  cluster-name="cluster3"  cluster-config="${jboss.server.config.dir}/modeshape/jgroups.xml"/>
```
* cluster name and server-internal stack name (for user migrating from 3.x):
```xml
<repository name="sample3"  cluster-name="cluster3" cluster-stack="udp"/>
```


modeshape-jcr/src/main/resources/org/modeshape/jcr/repository-config-schema.json
```diff
@@ -553,6 +553,28 @@
                 },
             }
         },
+        "clustering": {
+            "type": "object",
+            "description": "The clustering configuration. If absent, clustering will be disabled",
+            "additionalProperties": true,
```
rhauch: What additional properties are allowed/supported? Perhaps the `additionalProperties` value should be `false`.

modeshape-jcr/src/test/java/org/modeshape/jcr/RepositoryConfigurationTest.java
```diff
@@ -408,6 +409,28 @@ public void shouldReadJournalingConfiguration() {
...
+        assertEquals(channelConfig, clusteringConfiguration.getConfiguration());
+    }
+    
+    @Test
+    public void shouldUseDefaultClusteringValues() throws Exception {
+        RepositoryConfiguration config = RepositoryConfiguration.read("{ \"clustering\" : {} }");
+        RepositoryConfiguration.Clustering clusteringConfiguration = config.getClustering();
+        assertEquals(Default.CLUSTER_NAME, clusteringConfiguration.getClusterName());
+        assertEquals(Default.CLUSTER_CONFIG, clusteringConfiguration.getConfiguration());
+    }
```
rhauch: Would it be useful to have a test that verifies that no clustering is used when the clustering nested document is *missing*?

modeshape-jcr/src/test/resources/config/cluster/repo-config-clustered-persistent-1.json
```diff
@@ -21,6 +21,9 @@
             "minimumBinarySizeInBytes": 4096
         }
     },
+    "clustering" : {
+        //use defaults
```
rhauch: I forget ... do we allow comments in JSON files? (Presumably, since this file must be used somewhere.)

rhauch:
Just to clarify, no clustering is used when the `clustering` nested document is *missing*, correct? IOW, by default the repository will not be clustered, and only if a `clustering` nested document is specified, either empty or populated with up to its two fields, in the configuration. Is that correct?

modeshape-jcr/src/main/resources/org/modeshape/jcr/repository-config-schema.json
hchiorean: makes sense

modeshape-jcr/src/test/java/org/modeshape/jcr/RepositoryConfigurationTest.java
hchiorean: I suppose;  an assert can be added to one of the other tests which don't use clustering

hchiorean:
>Just to clarify, no clustering is used when the clustering nested document is missing, correct?

yes, when the `clustering` section is missing altogether no clustering is used.

>IOW, by default the repository will not be clustered, and only if a clustering nested document is specified, either empty or populated with up to its two fields, in the configuration. Is that correct?

correct

hchiorean:
amended previous commit with an additional unit test and the schema changes

rhauch:
Thanks. Looks great.

PR #1496, opened 2015-12-24
MODE-2546 Implements repository level locking

hchiorean:
With this PR all locking logic is removed from ISPN (even though ISPN is still used) in order to validate that the new approach works (i.e. the unit tests pass).

With the new approach the repository, via a `LockingService` abstraction together with the `LocalDocumentStore`, locks the nodes prior to writing changes as part of a transaction. The `unlock` operation is performed once the transaction completes (either successfully or not) via a transaction synchronization.

Similar to the way ISPN works, the new locking logic uses a `timeout` concept so that any given thread/node in a cluster can wait a given amount of time prior to timing out. The default timeout is 10 seconds (similar to ISPN) but can also be configured via the JSON and WF configurations:
```javascript 
{
  "name" : "Thorough",
  "lockTimeoutMillis" : 100
}
```
```xml
 <repository  name="sample" lock-timeout-millis="1000">
```

This configuration allows highly concurrent use cases to tweak the value and increase the timeout to avoid `TimeoutExceptions`


rhauch:
Does this lock service work only in a local, non-clustered environment? What kind of performance change is there?

modeshape-jcr/src/main/java/org/modeshape/jcr/locking/StandaloneLockingService.java
```diff
@@ -0,0 +1,222 @@
...
+                return false;
+            }
+
+            // we now have a lock but we need to make sure it hasn't just been marked for cleanup
+            // this may've been unlocked by another thread, so refresh to make sure a strong reference is active
+            validateLock(name, lock);
+
+            LOGGER.debug("{0} locked successfully", lock);
+            successfullyLocked.add(name);
+        }
```
rhauch: As mentioned in [line comments](https://github.com/hchiorean/modeshape/commit/c96d4ebfd2c8a18df13f3a433a8a2e2d4b092703#commitcomment-15155073) on one of the commits, I think this method should:

1) hold onto the acquired locks via a set (see how the `ClusteringService` does it in this same PR)
1) if any lock cannot be acquired, any already-acquired locks can be unlocked directly via this set, saving the need to look up the locks by name again on [line 103](https://github.com/ModeShape/modeshape/pull/1496/files#diff-5ba93ccae783c9c45dec04404e62e328R103)

In the other commit's comments, I suggested moving the `validateLock` methods into the `NamedLock`, but that won't necessarily be cleaner. So leaving those methods where they are is fine with me.

modeshape-jcr/src/main/java/org/modeshape/jcr/locking/StandaloneLockingService.java
```diff
@@ -0,0 +1,222 @@
...
+        return Arrays.stream(names)
+                     .map(this::unlock)
+                     .filter(Optional::isPresent)
+                     .map(Optional::get)
+                     .collect(Collectors.toList());
+    }
+    
+    private Optional<String> unlock(String name) {
+        LOGGER.debug("Attempting to unlock {0}", name);
+        NamedLock lock = locksByName.get(name);
```
rhauch: Shouldn't this synchronize on the `locksByName`?

modeshape-jcr/src/main/java/org/modeshape/jcr/locking/StandaloneLockingService.java
```diff
@@ -0,0 +1,222 @@
...
+        if (lock.isLocked()) {
+            LOGGER.debug("Failed to unlock {0} ", lock);
+            return Optional.of(name);
+        } else {
+            LOGGER.debug("{0} unlocked", lock);
+            return Optional.empty();
+        }
+    }
+
+    private class NamedLock implements Lock {
```
rhauch: I'd mark this as `@ThreadSafe` since it is potentially used concurrently from multiple threads to lock and/or unlock it, and so the `name` field is kept to be modified atomically.

rhauch:
There are only a few suggestions (see [this comment](https://github.com/ModeShape/modeshape/pull/1496#discussion_r48420511) and [this comment](https://github.com/ModeShape/modeshape/pull/1496#discussion_r48420564)) and one potential problem (see [this comment](https://github.com/ModeShape/modeshape/pull/1496#discussion_r48420524)). 

Everything else looks great!

hchiorean:
>How does this change affect performance?

looking at the overall build time, this change seems to have improved overall performance (still running ISPN but without explicit locking via code)

@rhauch I've amended the commit and actually changed the entire `StandaloneLockingService` implementation. There is a much simpler way to do this using a `ConcurrentHashMap` and still avoid OOM error by accumulating locks. The only trick was to remove a  lock from the map only if it's uncontended

modeshape-jcr/src/main/java/org/modeshape/jcr/locking/StandaloneLockingService.java
```diff
@@ -0,0 +1,123 @@
...
+            LOGGER.debug("Lock '{0}' is still locked with hold count '{1}'", name, lock.getHoldCount());
+            return Optional.of(name);
+        } else {
+            LOGGER.debug("'{0}' unlocked", name);
+            // if there is no contention for this lock, remove it
+            // otherwise leave it because the last queued thread will remove it once it completes
+            if (!lock.hasQueuedThreads()) {
+                locksByName.remove(name);
+            }
+            return Optional.empty();
```
rhauch: What happens if the lock is not locked on line 110, but by the time it gets here it has been locked again before we reach line 117? If there was only one other thread that acquired the lock, then the # of queued threads would be zero and we'd remove it from the map even though another thread is blocked on line 69 above attempting to lock the *existing* lock. (It would eventually unblock and acquire the lock, but that lock would no longer be in the map. That'd be bad.)

Also, it's not clear why you are returning a non-empty optional on line 112, since it might be possible that the lock was successfully unlocked for this thread by immediately acquired by another thread. So the return logic in this method seems to return a name when the named lock is still locked by somebody, even if it were successfully unlocked by me and I no longer hold any of the named locks. OTOH, the [JavaDoc for `LockingService.unlock(String...names)`](https://github.com/ModeShape/modeshape/pull/1496/files#diff-f872a09a1f66c182f534be128e5fa5aeR60) seems to imply that method returns only the names of those locks held by this thread that could not be released/unlocked -- essentially what of *my locks* could not be unlocked (which should only happen because unlock was called by a thread that did not acquire the lock).

How about the following logic, which I think fixes both problems?

    private Optional<String> unlock(String name) {
        LOGGER.debug("Attempting to unlock {0}", name);
        AtomicBoolean unlocked = new AtomicBoolean(false);
        locksByName.computeIfPresent(name, (key,lock)->{
            // Try to unlock the lock ...
            try {
                lock.unlock();
                unlocked.set(true);
            } catch (IllegalMonitorStateException e) {
                LOGGER.warn(JcrI18n.warnAnotherThreadAttemptingToUnlock, Thread.currentThread().getName(), lock);
                return lock;
            }
            if (lock.isLocked()) {
                LOGGER.debug("Lock '{0}' is still locked with hold count '{1}'", name, lock.getHoldCount());
                return lock;
            } else if ( lock.hasQueuedThreads() ) {
                LOGGER.debug("Lock '{0}' is not currently locked by will be", name);
                return lock;
            }
            // The lock is not used, and returning null will remove it from the map ...
            return null;
        });
        // A non-null reference to the lock means the lock is still there ...
        return unlocked.get() ? Optional.empty() : Optional.of(name);
    }

This does all of the work on the lock atomically with respect to the concurrent map, meaning it should not be possible for a race condition to occur between locking and unlocking a single lock. This logic is equivalent to your logic, except that it returns the name only when this thread's lock could not be released, which sounds like it better satisfies the JavaDoc intent.

What do you think?

hchiorean: I agree; the above are good points with regards to the fact that the logic from `unlock` should be performed atomically (hence the `computeIfPresent` method).

hchiorean:
@rhauch I've amended the commit with the suggested change, with the exception of the place where the `unlocked` variable is set (as per my previous comment)

rhauch:
@hchiorean: I understand that the lock should be removed only when it is uncontended, and the code does this via the return: if `null` is returned, the entry with the lock is removed from the map; otherwise, the entry is kept. 

However, the result of the `unlock` methods have different semantics, which IMO should include in the returned collection the name of *my* locks that *I* could not unlock. This is useful because it tells me what my method call was at least partially unsuccessful and that something else needs to be done. For example, perhaps the method can be tried again, or perhaps it signals that the `unlock` was called on a different thread than the `lock`, which signals a coding error.

OTOH, the current code returns the names of the locks that are still held by others, and I don't see how the caller of `unlock` can do anything with that knowledge because it doesn't distinguish an error condition from normal high-contention cases.

I still think the `unlocked.set(true)` should be called within the try block after the lock is successfully unlocked.

hchiorean:
>I still think the unlocked.set(true) should be called within the try block after the lock is successfully unlocked.

The `try` block may not necessarily successfully release the lock, even if the `lock.unlock` call doesn't throw an exception. This is because if a thread T1 calls `lock.lock` 3 times on the same lock instance, it has to symmetrically call `lock.unlock` 3 times on the same lock instance before that lock is actually unlocked.

Therefore, setting `unlock` to true would not be correct unless the holder count is 0.

rhauch:
Firstly, does ModeShape code lock a node more than once? And even if we do that or want to allow that, doesn't our code unlock all of the nodes it locked within a transaction? If so, then unlock should release all of the locks held by the caller, regardless of how many times it was locked. (BTW, what are the JGroups locking service semantics w/r/t locking multiple times and unlocking such locks?)

I didn't realize we allowed locking a node more than once, so I removed the use of `getHolderCount()` from my proposal. If we want to support unlocking a lock that was acquired multiple times, then we need to use `getHolderCount()` accordingly to either remove all locks held by the current thread (if that's the behavior we want) or to properly set the `unlocked` state.

rhauch:
@hchiorean: I probably wasn't terribly clear above. t would totally understand allowing a thread to lock a particular node more than once simply out of convenience so the caller doesn't have to track what locks they actually acquired. However, it also makes a lot of sense to me that unlocking a node releases *all* of my locks on that node. That keeps it simple.

Of course, it probably needs to match the semantics of the JGroups locking service.

hchiorean:
What we want in general is symmetrical handling of lock/unlock operations. Each lock should have a corresponding unlock.

JGroups (used when clustering) behaves differently than `ReentrantLock` used here: JGroups locks only once and unlocks only once, ignoring all other "reentrant" calls on the same lock instance (this is similar to your last comment)

`ReentrantLock` (used here for standalone) behaves differently: if a thread locks the same lock multiple times, it will increase the holder count. Therefore, the same number of times an unlock operation has to be called.

Locking the same `key` multiple times is actually a very common use case when you consider user transactions: within the same user transaction multiple `session.save` calls can occur and each `session.save` will lock the nodes it operated on. On each session.save however,  a transaction synchronization will be registered so that the corresponding locked nodes are always unlocked at the end of the transaction. This ensures the symmetry mentioned above.

hchiorean:
for simplicity however, I can look at changing this implementation to avoid the default `ReentrantLock` "reentrancy" contract

rhauch:
Right, so it seems like even though the `StandaloneLockingService` uses `ReentrantLock` internally, it should have the same semantics of the JGroups locking service. Otherwise, session write code will behave differently in standalone vs clustered.

> Locking the same key multiple times is actually a very common use case when you consider user transactions: within the same user transaction multiple session.save calls can occur and each session.save will lock the nodes it operated on. On each session.save however, a transaction synchronization will be registered so that the corresponding locked nodes are always unlocked at the end of the transaction. This ensures the symmetry mentioned above.

Totally agree that locking multiple times is a common use case. But *unlocking* all nodes happens just once, whether within an internal transaction (with a single session.save) or within a user transaction (with multiple session.saves).

rhauch:
> for simplicity however, I can look at changing this implementation to avoid the default ReentrantLock "reentrancy" contract

I think that'd be best, but IMO we should just loop within the unlock while the holder count is not zero. I can't think of an alternative JDK `Lock` implementation that we can use (we need the reentrancy), and looping is far easier than writing our own lock.

rhauch:
BTW, despite my critical comments, this is absolutely awesome work. I love the new design.

hchiorean:
> But unlocking all nodes happens just once, whether within an internal transaction (with a single session.save) or within a user transaction (with multiple session.saves).

Not quite: if thread T1 calls `session.save` 3 times and each session changes `key`, 2 things will happen:
1. `lockingService.lock(key)` will be called 3 times, each time reporting "true" (i.e. lock acquired)
2. 3 tx synchronizers will be registered with the transaction, each doing `lockingService.unlock(key)`

The difference in behavior (JGroups vs current standalone implementation) is that JGroups will release the lock faster (on the first call, since no reentrancy is used) while the standalone service will release the lock when the last tx synchronizer completes.

Both cases are correct, because the tx has been committed/rolled back so either the changes succeeded or not (in the standalone case however, there is some logical redundancy)

rhauch:
Okay, I agree there are some differences in regard to the multiple synchronizers. But we agree that both standalone and clustered lock services should work the same way, right? Personally, I think it's perfectly fine for the first unlock to release all locks, and for subsequent unlocks (if there are any) to do nothing. A simple while loop in the `computeIfPresent` function used in the private `unlock` would accomplish this. Of course, if you can think of a simpler way to do it, then great.

hchiorean:
+1, I agree that the implementations should behave the same

hchiorean:
@rhauch I've amended the commit again, but the logic of `unlock` is still a bit different, because now `IllegalMonitorStateExceptions` cannot really occur and have to be checked beforehand.

rhauch:
@hchiorean: Looks great! Very nice work.

PR #1499, opened 2016-01-05
MODE-2549 Adds PDF sequencer

RobSis:
Initial implementation of PDF sequencer based on Apache PDFBox.


sequencers/modeshape-sequencer-pdf/src/main/java/org/modeshape/sequencer/pdf/PdfBasicMetadata.java
```diff
@@ -0,0 +1,150 @@
...
+                keywords = metadata.getKeywords();
+                modificationDate = metadata.getModificationDate();
+                producer = metadata.getProducer();
+                subject = metadata.getSubject();
+                title = metadata.getTitle();
+
+                document.close();
+            }
+            return true;
+        } catch (IOException | IllegalArgumentException e) {
```
hchiorean: these should either 
a) be logged here and therefore the `logger` object from the parent sequencer passed in and used here
b) left to bubble up and caught and logged by the sequencer

in either case we want to be aware that these exceptions occurred.

sequencers/modeshape-sequencer-pdf/src/main/java/org/modeshape/sequencer/pdf/PdfBasicMetadata.java
```diff
@@ -0,0 +1,150 @@
...
+    public String getVersion() {
+        return version;
+    }
+
+    /**
+     * Set the input stream to the argument stream (or file).
+     *
+     * @param inputStream the input stream to read from
+     */
+    public void setInput( InputStream inputStream ) {
```
hchiorean: this seems to be always required, so it should be converted to a constructor

sequencers/modeshape-sequencer-pdf/src/main/resources/org/modeshape/sequencer/pdf/pdf.cnd
```diff
@@ -0,0 +1,57 @@
...
+  - pdf:creationDate (date)
+  - pdf:creator (string)
+  - pdf:keywords (string)
+  - pdf:modificationDate (date)
+  - pdf:producer (string)
+  - pdf:subject (string)
+  - pdf:title (string)
+  + pdf:xmp (pdf:xmp) = pdf:xmp
+  + pdf:attachment = pdf:xmp
+  + pdf:page = pdf:xmp
```
maschmid: Note, the current code doesn't seem to be creating the pdf:page and pdf:attachment nodes.

hchiorean:

RobSis:
Thanks for the input. 

hchiorean:

RobSis:
Hello. Yes! 

Thank you for notice!

PR #1500, opened 2016-01-05
MODE-2550 Adds OpenDocument Format sequencer

RobSis:
Initial implementation of ODF sequencer based on ODFDOM library.


sequencers/modeshape-sequencer-odf/src/main/java/org/modeshape/sequencer/odf/OdfMetadata.java
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+            if (doc instanceof SpreadsheetDocument) {
+                SpreadsheetDocument spreadsheet = (SpreadsheetDocument) doc;
+                sheets = spreadsheet.getSheetCount();
+            }
+
+            return true;
+        } catch (Exception e) {
+            // ignore
```
hchiorean: this should at least be logged so that we're aware something went wrong

hchiorean:
see [this comment](https://github.com/ModeShape/modeshape/pull/1499#commitcomment-15260782) on things that are still required for this to be included in the ModeShape distribution.

sequencers/modeshape-sequencer-odf/src/main/java/org/modeshape/sequencer/odf/OdfMetadata.java
```diff
@@ -0,0 +1,185 @@
...
+    public String getSubject() {
+        return subject;
+    }
+
+    /**
+     * Set the input stream to the argument stream (or file).
+     *
+     * @param inputStream the input stream to read from
+     */
+    public void setInput( InputStream inputStream ) {
```
hchiorean: this seems to be always required, so it should be converted to a constructor

hchiorean:

RobSis:
Thanks for the input. 

hchiorean:

RobSis:
Hello. Yes! 

Thank you for notice!

PR #1501, opened 2016-01-05
MODE-2551 Adds EPUB sequencer

RobSis:
Initial implementation of EPUB sequencer.


sequencers/modeshape-sequencer-epub/src/main/java/org/modeshape/sequencer/epub/EpubMetadata.java
```diff
@@ -0,0 +1,225 @@
...
+                            date = node.getTextContent();
+                        }
+                    }
+                }
+            }
+            zip.close();
+
+            return true;
+        } catch (Exception e) {
+            // log
```
hchiorean: the comment is correct, this should be logged or left to bubble up and logged in the sequencer

sequencers/modeshape-sequencer-epub/src/main/java/org/modeshape/sequencer/epub/EpubMetadata.java
```diff
@@ -0,0 +1,225 @@
...
+
+    private InputStream in;
+    private DataInput din;
+
+    /*
+     * Check that given file is supported by this sequencer.
+     */
+    public boolean check() {
+        try {
+            ZipInputStream zip = new ZipInputStream(in);
```
hchiorean: shouldn't this use try-with-resources ?

sequencers/modeshape-sequencer-epub/src/main/java/org/modeshape/sequencer/epub/EpubMetadata.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+
+    /**
+     * Set the input stream to the argument stream (or file). Note that {@link java.io.RandomAccessFile} implements
+     * {@link java.io.DataInput}.
+     *
+     * @param dataInput the input stream to read from
+     */
+    public void setInput( DataInput dataInput ) {
```
hchiorean: is this used anywhere ? it seems only the `setInputStream(InputStream)` is used

hchiorean:
The `EpubMetadata` seems to only be used together with an `InputStream`. As such,
a) its constructor should be changed and require this input stream (and hence no more setter)
b) the `DataInput` member should be removed if not used

Also, see [this comment](https://github.com/ModeShape/modeshape/pull/1499#commitcomment-15260782) on things that are still required for this to be included in the ModeShape distribution.

sequencers/modeshape-sequencer-epub/src/main/java/org/modeshape/sequencer/epub/EpubMetadata.java
```diff
@@ -0,0 +1,225 @@
...
+
+    /*
+     * Check that given file is supported by this sequencer.
+     */
+    public boolean check() {
+        try {
+            ZipInputStream zip = new ZipInputStream(in);
+            ZipEntry entry = null;
+            while ((entry = zip.getNextEntry()) != null) {
+                if (entry.getName().endsWith("/content.opf")) {
```
maschmid: I don't believe the EPUB spec requires the rootfile to be named "content.opf"

The name of the rootfile (or rootfiles) is definied in META-INF/container.xml  (see http://www.idpf.org/epub/301/spec/epub-ocf.html#sec-container-metainf-container.xml )

hchiorean:

RobSis:
Thanks for the input. 

hchiorean:

RobSis:
Hello. Yes! 

Thank you for notice!

PR #1504, opened 2016-01-05
MODE-2450 Fixes the handling of corrupted locks from the jcr:system area which are still being referenced by the parent node

hchiorean:


modeshape-jcr/src/main/java/org/modeshape/jcr/RepositoryLockManager.java
```diff
@@ -131,14 +132,21 @@ protected void refreshFromSystem() {
...
             }
             if (mutableLocks != null) {
                 system.save();
+                for (Iterator<Map.Entry<NodeKey, ModeShapeLock>> locksIterator = locksByNodeKey.entrySet().iterator(); 
+                     locksIterator.hasNext();) {
+                    NodeKey lockKey = locksIterator.next().getValue().getLockKey();
+                    if (corruptedLocks.contains(lockKey)) {
+                        locksIterator.remove();
+                    }
+                }
```
rhauch: This block is done after the save. Is that correct?

hchiorean: yes, its purpose is to clear any potential in-memory locks stored in the `locksByNodeKey` member.

rhauch:
This looks good.

PR #1507, opened 2016-01-07
MODE-2518: Attach binary content jcr:content

okulikov:
This PR fixes bug when property with null value (ie just created node) is not displayed by property editor.

### DIFF
--- a/web/modeshape-web-explorer/src/main/java/org/modeshape/web/server/JcrServiceImpl.java
+++ b/web/modeshape-web-explorer/src/main/java/org/modeshape/web/server/JcrServiceImpl.java
@@ -352,15 +352,16 @@ public class JcrServiceImpl extends RemoteServiceServlet implements JcrService {
             String name = def.getName();
             String type = PropertyType.nameFromValue(def.getRequiredType());
             
-            Property p;
+            Property p = null;
             try {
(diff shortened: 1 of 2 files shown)


web/modeshape-web-explorer/src/main/java/org/modeshape/web/server/JcrServiceImpl.java
```diff
@@ -352,15 +352,16 @@ private AccessControlList findAccessList( AccessControlManager acm,
...
             
-            Property p;
+            Property p = null;
             try {
                 p = node.getProperty(def.getName());
-                String display = values(def, p);
-                String value = p.isMultiple() ? multiValue(p)
-                        : singleValue(p, def, repository, workspace, path);
-                list.add(new JcrProperty(name, type, value, display));
             } catch (PathNotFoundException e) {
```
hchiorean: should this exception be ignored ?

okulikov: yes. it means that property is not assign yet.

hchiorean:

rhauch:
@hchiorean, looks good.

PR #1508, opened 2016-01-07
MODE-2534 Teiid DDL sequencer doesn't support GLOBAL or LOCAL TEMP TABLE options

elvisisking:
Implemented the processing of DDL that creates local and foreign temporary tables. This included adding new node types for all types of temporary tables.

**I am still unable to get a good local build as the cluster repository tests are failing in `modeshape-jcr`. The changed files are only in `modeshape-sequencer-ddl`.**


hchiorean:
looks good to me; @rhauch  ?

rhauch:
@hchiorean, I agree.

PR #1511, opened 2016-01-11
MODE-2555 Adds the support of expressions to the <node-types/> WF configuration element

hchiorean:
This also introduces a more generic ability for a repository configured in WF to use the CL of another external (dependant) module when resolving configuration elements such as initial content files and node type files. 
For example, configuring a repository as:
```xml
<repository name="externalDependenciesRepository" depends-on="deployment.external-dependencies.war"/>
```
means that the repository will be able to locate & load node-types and initial content files located in the `external-dependencies.war` webapp


rhauch:
Looks good.

PR #1512, opened 2016-01-11
MODE-2508 Add OPTIONS capability to Procedure ResultSet

elvisisking:
- changed parser to handle options for result sets.
- added parser and sequencer tests.


rhauch:
Looks good.

PR #1513, opened 2016-01-11
MODE-2554 Add Geometry Type to Teiid DDL dialect

elvisisking:
Just needed to add the new GEOMETRY data type to the data type enum. No additional code needed. Wrote test to make sure the parser recognized the new data type.


hchiorean:
looks good

PR #1514, opened 2016-01-13
MODE-2499: Executing queries from ModeShape Explorer doesn't validate dialects

okulikov:
This PR adds one more syntax checking rule for  the `XPathParser`: 
- after parsing no more tokens should remain. 

Without this check the beginning of some JCR-SQL2 statements (or even random text) will be interpreted as xpath query while remainder will be ignored. 

Web-Explorer correctly shows exception when it is fired by MS.


hchiorean:
@okulikov: this looks good, but I have an additional question: what does the `search` option (from `Query Language` dropdown) mean ?

okulikov:
@hchiorean , it comes from `QueryManager.getSupportedQueryLanguages()`

hchiorean:
@okulikov: right, guess it's this one:
```java
    /**
     * A string constant representing the ModeShape full-text search query language.
     */
    public static final String FULL_TEXT_SEARCH = "search";
```

hchiorean:

rhauch:
@okulikov, @hchiorean: Looks good to me.

PR #1516, opened 2016-01-14
MODE-2560 Fixes the behavior of copying versionable nodes

hchiorean:


rhauch:
@hchiorean: this change looks good. What was the core fault, and did it result in a repository state that should be corrected before the code works properly?

hchiorean:
@rhauch: there were a couple of faults:
1. when performing a `workspace.copy` operation none of the versionable properties on the node (which are essentially references) should be copied across. As per the spec, the copied node should get its own base version, version history etc
2. the `copiedFrom` property on the copied node's version history should be a `weakreference` (as per the spec) but the previous code created a strong reference. 

These 2 together meant that there were some strong references towards a version node which made it impossible to remove.

>it result in a repository state that should be corrected before the code works properly?

not sure what you mean by this. If any code prior to this fix performed a `workspace.copy` on a versionable node, (1) from above means that the references across version histories are corrupted and there is not way to fix that as it influence the way the version history was created for the copied node

rhauch:
@hchiorean: What do we expect users to do if they already have version histories with this wrong content?

hchiorean:
drop their data :)
this may not sound helpful, but the references are permanently corrupted and there is no way to tell "good version histories" from "bad version histories". 
If the problem was only (2) from above, then this could be perhaps fixed automatically, but (1) makes this impossible IMO.

rhauch:
Regardless, this change looks good.

PR #1517, opened 2016-01-15
MODE-2559 Updates the backup & restore so that if the "compress" attribute is set, it also compresses binary values

hchiorean:


modeshape-common/src/main/java/org/modeshape/common/util/FileUtil.java
```diff
@@ -318,6 +318,20 @@ public static void addFileToZip(String path, String srcFile, ZipOutputStream zip
...
+    /**
+     * Returns the extension of a file, including the dot.
+     * 
+     * @param filename the name of a file, may be not be null
+     * @return the file extension or an empty string if the extension cannot be determined.
+     */
+    public static String getExtension(final String filename) {
+        if (filename == null) {
+            throw new IllegalArgumentException("filename cannot be null");
+        }
```
rhauch: This is perfectly fine, but ICYMI also and for future reference, with Java 8 we can now do this using the built-in `java.util.Objects` class:

    Objects.requireNotNull(filename,"filename cannot be null");

Even better is that the methods return the parameter, so you can use it in constructors and method params quite easily to check and assign with a one-liner.

rhauch:
Looks great.

modeshape-common/src/main/java/org/modeshape/common/util/FileUtil.java
hchiorean: this is against 4.x, so no Java 8
otherwise +1

rhauch: I corrected my statement ... it arrived in Java 7.

rhauch: And again, I'm not suggesting you change this PR. Looks great.

hchiorean:
amended the commit; I could've sworn `Objects.` was added only in JDK 8 :smile:

PR #1519, opened 2016-02-03
MODE-2565 - Added test to demonstrate lucene re-indexing issue.

lucasweb78:
The test will not fail, but if you look at the log output you will see that after restarting the repository the entire repository is re-indexed again even though there have be no changes since the repository was stopped.


hchiorean:
the reported issue is valid, but this test case is not. 

The repository configuration used in this test case _does not store data anywhere_ (which includes index definitions). As such, each restart is the equivalent to a full system area re-initialization, creating and storing index definitions etc. So it's perfectly normal for the indexes to grow as new system area data is created.

PR #1520, opened 2016-02-15
MODE-2527-2528 Removes any ISPN dependencies and creates new persistence abstraction

hchiorean:
This is a "major" PR for the `master` branch which removes ISPN and changes ModeShape's persistence mechanism to use a RDBMS storage. 

All the code is fully functional and *all* the unit tests (including clustering) pass using the default configuration which is H2 in-memory.

The key points about this PR are:
* [modeshape-schematic](https://github.com/hchiorean/modeshape/tree/MODE-2527-2528/modeshape-schematic) contains only the BSON and schema related code, together with a high-level persistence abstraction which is [SchematicDB](https://github.com/hchiorean/modeshape/blob/MODE-2527-2528/modeshape-schematic/src/main/java/org/modeshape/schematic/SchematicDb.java). The package name has been changed to `org.modeshape.*` and all ISPN references removed. Like before, this module _doesn't have any other ModeShape-specific dependencies_

* [persistence](https://github.com/hchiorean/modeshape/tree/MODE-2527-2528/persistence) is a new top-level module which will host the different persistence provider implementations and depends only on `modeshape-schematic` and `modeshape-common`.

* each persistence implementation integrates at runtime via a `ServiceProvider` mechanism, processed by [Schematic](https://github.com/hchiorean/modeshape/blob/MODE-2527-2528/modeshape-schematic/src/main/java/org/modeshape/schematic/Schematic.java#L66)

* [modeshape-persistence-relational](https://github.com/hchiorean/modeshape/tree/MODE-2527-2528/persistence/modeshape-persistence-relational) is the first concrete `SchematicDB` implementation which stores data in RDBMS. The ones I've smoke tested are: MySQL 5.x, Postgres 9, Oracle 11g and H2 (the default)

*  the new relational provider integrates with the repository configuration like:
```javascript
 {
    "name" : "Clustered Repository with Incremental Indexes",
    "storage" : {
        "persistence" : {
            "type" : "db",
            "connectionUrl": "jdbc:h2:file:./target/clustered/db;AUTO_SERVER=TRUE"
        }
    }
}
```

and 
```xml
<subsystem xmlns="urn:jboss:domain:modeshape:3.0">
    <repository name="sample">
       <db-persistence url="jdbc:h2:mem:modeshape;DB_CLOSE_DELAY=0;" compress="false" create-on-start="false" datasource-jndi="name"
                       driver="org.h2.Driver" drop-on-exit="false" fetch-size="1000" password="s" pool-size="11" table-name="MODESHAPE_REPOSITORY"
                       username="sa"/>
    </repository>
</subsystem>
```
* without an explicit configuration, H2 in memory is used

* most of the ModeShape unit tests use the above, with some exceptions (for example clustering) where H2 file storage is used. With this in mind, the overall performance judged by the total build time and some read-write performance tests is comparable (probably a 5% decrease) to ModeShape 4.x which uses most of the time Infinispan in-memory (i.e. just a `ConcurrentHashMap`).  This is decent IMO, considering that the new code _always goes through JDBC_


rhauch:
@hchiorean: See above for a few questions, but overall this is really outstanding work. All of the changes are clear, thorough, and very-well written.

I have one other suggestion: document in the codebase's README.md file that Docker is now require to test MySQL and Oracle, and the commands that should be used to run the MySQL and Oracle Docker containers, since the build doesn't do this automatically.

hchiorean:
>document in the codebase's README.md file that Docker is now require to test MySQL and Oracle, and the commands that should be used to run the MySQL and Oracle Docker containers, since the build doesn't do this automatically.

the old `mysql`, `oracle`, `postgres` etc profiles are still present. So `Docker` isn't a requirement atm; I just added some additional `_docker` profiles because it was the quickest option for testing on my machine.

As you pointed out previously, the `_docker` profiles (which I'll probably rename to `_local_docker`) are just for my machine. I'll open a JIRA for adding "automated docker" tests after which the `Readme` can be updated.

wdyt ?

hchiorean:
added a new commit which makes some changes based on the above comments; it also fixes the locking logic of the `ClusteringService`.
I've renamed the  `_docker` profiles to `_local_docker` which hopefully expresses the intent better.

hchiorean:
@rhauch  I've amended the commit and changed the logic to use `computeIfAbsent` which should be atomic with respect to another thread removing the same lock

rhauch:
@hchiorean: It still seems like it will not handle the case I mentioned above, where the lock already exists in the map when `computeIfAbsent(...)` is called but it is removed before the code tries to lock it on [line 238](https://github.com/hchiorean/modeshape/commit/7cbda8f040f0dff315025c8f38908bae73041851#diff-7a3817f69891df411ce49d5aee1ead1bR238) (outside the function called by `computeIfAbsent`). Wouldn't it be better to use `compute(...)` and always try locking within the function?

hchiorean:
@rhauch: changed the logic to use `compute`

rhauch:
@hchiorean: thanks for addressing this. I think this will behave much better in the pathological cases. The PR looks good to me. Again, great work!

PR #1521, opened 2016-02-17
MODE-2565 Fixes the behavior of the Lucene index provider to avoid reindexing at startup if data exists

hchiorean:
Also, it updates the general IF_MISSING reindexing logic, so that only the indexes which actually do require reindexing are reindexed, as opposed to the all the indexes of a particular provider.


rhauch:
Looks good.

PR #1524, opened 2016-02-19
MODE-2567 - Added a test and a fix to ensure that the Lucene index provider escapes Ampersands

lucasweb78:
Lucene Index Provider does not correctly handle Like constraints containing an Ampersand with a leading or trailing space


lucasweb78:
Added a second commit which fixes the issue.

hchiorean:
squashed and merged in `master`, thanks @lucasweb78

PR #1525, opened 2016-02-23
MODE-2570 Trims the modeshape binary distribution by removing the demos, test and test-source artifacts.

hchiorean:
It also updates the JavaDoc plugin number and configuration to run with JDK 8.


rhauch:
Looks good.

PR #1526, opened 2016-02-26
MODE-2573 - Using OFFSET on queries which use indexes loops over the same first 100 nodes

lucasweb78:
Test Case demonstrating issue.


hchiorean:
superseded by https://github.com/ModeShape/modeshape/pull/1533

PR #1527, opened 2016-02-29
MODE-2528 Optimizes some of the RelationalDB provider logic to bulk load nodes after locking

hchiorean:
It also adds a way to track new nodes per transaction in order to avoid unnecessary DB hits


persistence/modeshape-persistence-relational/src/main/java/org/modeshape/persistence/relational/StatementsProvider.java
```diff
@@ -104,6 +108,26 @@ protected Void dropTable(Connection connection) throws SQLException {
...
+                    throw new RelationalProviderException(e);
+                }
+            });
+            try (ResultSet rs = ps.executeQuery()) {
+                return function.apply(rs);
+            }
+        }
+    }
+    
+
```
rhauch: Is there a limit to the number of IDs that a single statement can handle? Should this method perhaps do no more than X identifiers at a time?

hchiorean: I'm not aware of a hard limit on the number of IDs in an IN clause

rhauch: * [MySQL](http://dev.mysql.com/doc/refman/5.1/en/comparison-operators.html#function_in): "limited by the max_allowed_packet value."
* [SQL Server](https://msdn.microsoft.com/en-us/library/ms143432.aspx): "limited by 65,536 * Network Packet Size" (which defaults to 4K).
* [Oracle](http://stackoverflow.com/questions/19002792/why-oracle-in-clause-has-limit-of-1000-only-for-static-data) has an expression limit of 1000 values.
* PostgreSQL [apparently](http://stackoverflow.com/questions/2426331/in-mysql-or-postgres-is-there-a-limit-to-the-size-of-an-in-1-2-n-statement) slows dramatically with lots of values.

Perhaps you could add a warnings to the JavaDoc and/or log stating that the method may not work with large numbers of IDs.

hchiorean:
amended the commit adding the aforementioned JavaDoc and also changing the backup logic to use batches when loading the keys from the DB in order to export them (because backing up is the place which is most likely to hit a potential problem rather than regular `session.saves` which should not operate on huge number of nodes at once)

rhauch:
Looks great.

PR #1528, opened 2016-03-02
MODE-2563 Ports the ModeShape AS kit to Wildfly 10.

hchiorean:
The current kit is also backwards-compatible to Wildfly 9, to it will still work with this version as well.


rhauch:
Looks good.

PR #1529, opened 2016-03-03
MODE-2543 Updates to jboss-parent 20 (JDK 8).

hchiorean:
It also makes some build changes removing/renaming some properties.


rhauch:
Looks good.

PR #1531, opened 2016-03-17
MODE-2569 Adds a new file-based persistence for ModeShape 5

hchiorean:
This PR adds a new fs (and in-memory) provider using H2's `MVStore`. 

For ModeShape's use case - where concurrent writers do not overlap (i.e. the repository exclusively locks each key) - this seems to work fine. Also, H2's `MVStore` supports transactions and is durable. In case of unexpected failures, it will rollback any uncommitted changes (it uses internally a tx log, so barring different bugs, it seems to be able to ensure data consistency)

The new file-based persistence is supported both in the JSON and WF kit:
```javascript
"persistence" : {
       "type" : "file",
       "path": "target/persistent_repository/db"
 }
```
or

```xml
<file-persistence path="${jboss.server.data.dir}/modeshape/artifacts"/>
```
In addition, this PR contains some commits related to other functionality from https://issues.jboss.org/browse/MODE-2525:
- https://issues.jboss.org/browse/MODE-2528: refactors some of the DB provider logic, making it easier in the future to extend the statements & JDBC logic for a particular DB type/version. Also, it changes the binary storage format to `Bson` because this seems to perform better in terms of serialization-deserialization throughput
- https://issues.jboss.org/browse/MODE-2546: fixes some concurrency bugs around the exclusive locking logic, where `ReentrantLock.hasQueuedThreads` is not safe to use.


deploy/jbossas/modeshape-jbossas-subsystem/src/main/resources/schema/modeshape_3_0.xsd
```diff
@@ -574,7 +583,25 @@
...
+
+    <xs:complexType name="file-persistence">
+        <xs:attribute name="path" type="xs:string" use="optional" default="${jboss.server.data.dir}/modeshape">
+            <xs:annotation>
+                <xs:documentation>
+                    The absolute path on the FS where the repository data should be persisted. If not specified,
+                    defaults to ${jboss.server.data.dir}/modeshape/${repository.name}
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
```
rhauch: Is it no longer idiomatic to define a path via something relative to another (which defaults to the data dir)? Doing that is more complicated, but it seems like it's far less fragile than using an absolute path.

hchiorean: i never understood where that idiom came from in the first place and it always seemed over complicated to me since AS has environment variables for all the server-specific folders :wink: 

if there's some sort of logical explanation for separating paths in 2 I'm happy to change it

rhauch: As long as the "path" values can contain those environment variables, then using a single "path" is far better. Thanks.

rhauch:
Looks good.

deploy/jbossas/modeshape-jbossas-subsystem/src/main/resources/schema/modeshape_3_0.xsd
hchiorean: the AS config gets translated by our AS services to JSON; this in turn does have support for env variables.

Considering that the [full list](https://docs.jboss.org/author/display/WFLY8/Command+line+parameters) of variables covers pretty much all of the "interesting" server dirs, it should be enough for our needs.

PR #1532, opened 2016-03-18
MODE-2584 Fixes the restoring of binary values which have to be marked as "used" after a restore is complete

hchiorean:


rhauch:
Looks good

PR #1534, opened 2016-03-21
MODE-2580 Removes the optimization added for MODE-1856 which caused erroneous PROPERTY_CHANGED_EVENTS

hchiorean:


rhauch:
Looks good.

PR #1538, opened 2016-03-22
MODE-1660: Initial implementation for command line shell

okulikov:
This PR provides implementation of the command line shell for initial review.  Communication with repository is build using JMX. JMX remote port is 9999 and it is hard coded at this stage. 

Console main class: org.modeshape.jcr.shell.ShellConsole. 
Immediately after start the only available command is 'list' which shows reachable repositories (MBeans). 
Command connect [-u user -p password -w workspace] Repository_name joins console with repository.
<Tab> - lists available commands for the current context.

### DIFF
--- a/modeshape-shell/pom.xml
+++ b/modeshape-shell/pom.xml
@@ -1,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.modeshape</groupId>
+        <artifactId>modeshape-parent</artifactId>
+        <version>5.1-SNAPSHOT</version>
+        <relativePath>../modeshape-parent/pom.xml</relativePath>
+    </parent>
+
+    <!-- The groupId and version values are inherited from parent -->
+    <artifactId>modeshape-shell</artifactId>
+    <packaging>jar</packaging>
+    <name>ModeShape Shell</name>
+    <description>ModeShape command line shell unit</description>
+    <url>http://www.modeshape.org</url>
+
+    <!--
+      Define the dependencies. Note that all version and scopes default to those defined in the dependencyManagement section of the
+      parent pom.
+    -->
+    <dependencies>
+        <dependency>
+            <groupId>org.jboss.aesh</groupId>
+            <artifactId>aesh</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.jcr</groupId>
+            <artifactId>jcr</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.modeshape</groupId>
+            <artifactId>modeshape-jcr-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.modeshape</groupId>
+            <artifactId>modeshape-jcr</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.modeshape</groupId>
+            <artifactId>modeshape-common</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.modeshape</groupId>
+            <artifactId>modeshape-schematic</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.sshd</groupId>
+            <artifactId>sshd-core</artifactId>
+        </dependency>
+    </dependencies>
+    <profiles>
+        <profile>
+            <id>assembly</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <artifactId>maven-assembly-plugin</artifactId>
+                        <configuration>
+                            <descriptorRefs>
+                                <descriptorRef>module-with-dependencies</descriptorRef>
+                            </descriptorRefs>
+                            <attach>false</attach>
+                            <finalName>${binary.dist.name}</finalName>
+                            <appendAssemblyId>false</appendAssemblyId>
+                        </configuration>
+                        <executions>
+                            <execution>
+                                <phase>package</phase>
+                                <goals>
+                                    <goal>single</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+</project>
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/ModeShapeEngine.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/ModeShapeEngine.java
@@ -22,6 +22,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.Map;
+import java.util.Properties;
 import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
@@ -44,6 +45,7 @@ import org.modeshape.common.util.CheckArg;
 import org.modeshape.common.util.ImmediateFuture;
 import org.modeshape.common.util.NamedThreadFactory;
 import org.modeshape.jcr.api.Repositories;
+import org.modeshape.shell.ShellServer;
 
 /**
  * A container for repositories.
@@ -59,6 +61,10 @@ public class ModeShapeEngine implements Repositories {
     private ExecutorService repositoryStarterService;
     private volatile State state = State.NOT_RUNNING;
 
+    //allows specify engine configuration
+    private Properties config;
+    private ShellServer sshServer;
+    
     public enum State {
         NOT_RUNNING,
         STARTING,
@@ -70,6 +76,16 @@ public class ModeShapeEngine implements Repositories {
     public ModeShapeEngine() {
     }
 
+    /**
+     * Creates engine with specific parameters.
+     * 
+     * MODE-1660: Enable specify shell server bootstrap parameters.
+     * @param config 
+     */
+    public ModeShapeEngine(Properties config) {
+        this.config = config;
+    }
+    
     protected final boolean checkRunning() {
         if (state == State.RUNNING) return true;
         throw new IllegalStateException(JcrI18n.engineIsNotRunning.text());
@@ -100,6 +116,13 @@ public class ModeShapeEngine implements Repositories {
             ThreadFactory threadFactory = new NamedThreadFactory("modeshape-start-repo");
             repositoryStarterService = Executors.newCachedThreadPool(threadFactory);
 
+            if (config != null && config.containsKey("ssh.server.enabled")) {
+                boolean enabled = Boolean.parseBoolean(config.getProperty("ssh.server.enabled"));
+                if (enabled) {
+                    sshServer = new ShellServer(this, config);
+                    sshServer.start();
+                }
+            }
             state = State.RUNNING;
         } catch (RuntimeException e) {
             state = State.NOT_RUNNING;
@@ -220,11 +243,17 @@ public class ModeShapeEngine implements Repositories {
                 repositoryStarterService.shutdown();
                 repositoryStarterService = null;
 
+//                this.unregisterMBean();
                 // Do not clear the set of repositories, so that restarting will work just fine ...
                 this.state = State.NOT_RUNNING;
             } else {
                 // Could not shut down all repositories, so keep running ..
                 this.state = State.RUNNING;
+            }
+            
+            //stop SSH server if it is exists
+            if (sshServer != null) {
+                sshServer.stop();
             }
         } catch (RuntimeException e) {
             this.state = State.RUNNING;
@@ -613,4 +642,5 @@ public class ModeShapeEngine implements Repositories {
             lock.unlock();
         }
     }
+    
 }
--- a/modeshape-parent/pom.xml
+++ b/modeshape-parent/pom.xml
@@ -209,6 +209,8 @@
         <version.org.wildfly.plugings.wildfly-maven-plugin>1.0.2.Final</version.org.wildfly.plugings.wildfly-maven-plugin>
         <version.elasticsearch>2.1.0</version.elasticsearch>
         <version.jna>4.1.0</version.jna>
+        <version.jboss.aesh>0.66.4</version.jboss.aesh>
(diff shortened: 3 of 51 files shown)


modeshape-jcr/pom.xml
```diff
@@ -262,6 +262,12 @@
             <groupId>org.apache.cassandra</groupId>
             <artifactId>cassandra-all</artifactId>
         </dependency>
+        <dependency>
```
hchiorean: this should be part o the parent POM; `jar` is the default type and not needed

modeshape-jcr/src/main/java/org/modeshape/jcr/JcrRepository.java
```diff
@@ -117,6 +117,8 @@
 import org.modeshape.jcr.security.EnvironmentAuthenticationProvider;
 import org.modeshape.jcr.security.JaasProvider;
 import org.modeshape.jcr.security.SecurityContext;
+import org.modeshape.jcr.shell.Interpreter;
+import org.modeshape.jcr.txn.SynchronizedTransactions;
```
hchiorean: why is this import required here ?

modeshape-jcr/src/main/java/org/modeshape/jcr/shell/LocalSession.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package org.modeshape.jcr.shell;
+
+import javax.jcr.Session;
+
+/**
+ *
+ * @author kulikov
+ */
+public class LocalSession {
```
hchiorean: is this wrapper required ? it doesn't seem to add anything else on top of a `javax.jcr.Session`

modeshape-jcr/src/main/java/org/modeshape/jcr/shell/Shell.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.rmi.RemoteException;
+import java.util.logging.Logger;
+
+/**
+ *
+ * @author kulikov
+ */
+public class Shell implements ShellMBean {
+    private Interpreter shell;
```
hchiorean: can this be `final` ?

modeshape-jcr/src/main/java/org/modeshape/jcr/shell/ShellConsole.java
```diff
@@ -0,0 +1,223 @@
...
+        }
+    }
+
+    private void scriptCommand(Command command, String fileName) {
+        if (fileName == null) {
+            console.getShell().err().println(command.help());
+            return;
+        }
+
+        try {
```
hchiorean: wouldn't a `try-with-resources` be better here ? It doesn't look like the streams are closed anywhere

modeshape-jcr/src/main/java/org/modeshape/jcr/shell/cmd/Command.java
```diff
@@ -0,0 +1,211 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.modeshape.jcr.shell.Interpreter;
+
+/**
+ *
+ * @author kulikov
+ */
+public abstract class Command {
```
hchiorean: can this class (or any of its subclasses) ever be used from multiple threads ? i.e. is thread safety an issue to consider or not (I'm asking because there seems to be a lot of state held)

modeshape-jcr/src/main/java/org/modeshape/jcr/shell/cmd/QueryCommand.java
```diff
@@ -0,0 +1,71 @@
...
+        builder.append("\n");
+        RowIterator it = res.getRows();
+        while (it.hasNext()) {
+            Row row = it.nextRow();
+            Value[] values = row.getValues();
+            for (int i = 0; i < values.length; i++) {
+                if (i > 0) builder.append("\t\t\t");
+                builder.append(values[i]);
+            }
+            builder.append("\n");
```
hchiorean: shouldn't this be `System.getProperty("line.separator")` ?

modeshape-jcr/src/main/java/org/modeshape/jcr/shell/Interpreter.java
```diff
@@ -0,0 +1,96 @@
...
+    
+    public JcrRepository getRepository() {
+        return repository;
+    }
+    
+    public Session getJcrSession() {
+        return session.jcrSession();
+    }
+    
+    public void setJcrSession(Session session) {
```
hchiorean: I'm not convinced this method should be here; i.e. I don't think it makes sense to pass a `session` from the outside. 

If I understand correctly, typical usages looks like:
```
shell.setJcrSession(shell.getRepository().login(workspace));
```
so even though this class has a reference to the `repository`, code is attempting to set a `session` by calling `login` on the repository 

IMO this should be encapsulated better and this class should have methods which get a new session instance without requiring it to be passed from the outside. For example:
```
login(String workspace)
login(String workspace, String username)
``` 
etc
which internally would set a new session instance.

modeshape-jcr/src/main/java/org/modeshape/jcr/shell/cmd/ConnectCommand.java
```diff
@@ -0,0 +1,68 @@
...
+            return String.format(RESPONSE, workspace, userName);
+        }
+        
+        shell.setJcrSession(shell.getRepository().login(creds));
+        return String.format(RESPONSE, "default", userName);
+    }
+
+    @Override
+    public String help() {
+        return "usage: connect OPTIONS Repository\n"
```
hchiorean: shouldn't the `help` messages in general come from `I18n` resources ?

hchiorean:
@okulikov, @rhauch: IMO this looks like a good start, but I would like to bring back a question from https://issues.jboss.org/browse/MODE-1660: shouldn't part of this be in a separate GitHub repository ?

The way I understand this console, there are two main units of work:

1. `console-server` - the part of the console code which adapts/exposes repository functionality for the client. This naturally has to be collocated with the main ModeShape codebase, but IMO _should be a separate Maven module_, not part of `modeshape-jcr`
2. `console-client` - the part of the console code with which clients interact directly to achieve a desired functionality. This part IMO should be separate (release and code-base wise) from the main ModeShape codebase, in a separate GitHub repository altogether

Please correct me if I'm wrong on any of the above 2

okulikov:
@hchiorean , agree with you. The thing is that second part is represented by _only one simple class_.  So it has minimal impact on first part.

Another preference to keep all parts together is shell use case. Irrespective of the remote capabilities of JMX it is more convenient to rich remote server using telnet or secure shell at the OS level and start console-client from the server-side deployment (just for instance, connect from mobile device). 

So it seems logical to hold both parts at the same development repository as well.

hchiorean:
perhaps, but we also need to keep in mind that the console is not something which is likely to ever be productized. As such, it needs to be independent/separate of the main modeshape artifacts.

So even if we don't have a separate GitHub repository for it, it definitively needs to be a separate Maven artifact which can be excluded/included in certain distributions.

rhauch:
@okulikov, @hchiorean: Won't the same console version work across multiple server releases? That seems to imply it should be a separate release/download/install.

okulikov:
@rhauch , as mentioned above console contains two parts. First part is a part of server while second part definitely will work across releases but it is single class. So it seems that the best way is to keep all pieces together and as @hchiorean suggested isolate into dedicated maven module.

rhauch:
@okulikov: If it's a separate module, then WDYT about the module's package name being `org.modeshape.shell` (rather than the current `org.modeshape.jcr.shell`)? 

And, is there a way for an admin to turn off the server-side of the shell to prevent all access this way?

okulikov:
@rhauch , agree

modeshape-jcr/src/test/java/org/modeshape/jcr/JcrRepositoryStartupTest.java
```diff
@@ -624,7 +624,7 @@ public void shouldRun4_0_0_Beta3_UpgradeFunction() throws Exception {
         }, config);
     }
 
-    @Test
+//    @Test
```
hchiorean: why is this commented out ?

modeshape-shell/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+            <artifactId>modeshape-jcr</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.modeshape</groupId>
+            <artifactId>modeshape-common</artifactId>
+        </dependency>
+    </dependencies>
+    <build>
+        <!-- resources>
```
hchiorean: if this is not required, it should be removed

modeshape-shell/src/main/java/org/modeshape/shell/Shell.java
```diff
@@ -0,0 +1,44 @@
...
+    private final Logger log = Logger.getLogger("Shell");
+    
+    public Shell(Repository repository) {
+        this.interpreter = new Interpreter(repository);
+    }
+    
+    @Override
+    public Message perform(Message cmd) throws RemoteException {
+        try {
+            log.severe("====== MBean executing command: " + cmd);
```
hchiorean: this looks like an informational message to me. Why is it logged as `severe` ?

modeshape-shell/src/test/java/org/modeshape/shell/cmd/ShellTest.java
```diff
@@ -0,0 +1,384 @@
...
+import static org.junit.Assert.assertTrue;
+import org.junit.Before;
+import org.modeshape.jcr.ModeShapeEngine;
+import org.modeshape.jcr.RepositoryConfiguration;
+import org.modeshape.jmx.RepositoryStatisticsMXBean;
+import org.modeshape.shell.Message;
+import org.modeshape.shell.ShellI18n;
+
+/**
+ * Unit test for {@link RepositoryStatisticsMXBean}
```
hchiorean: I suspect this description should read something else :)

modeshape-shell/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+            <groupId>javax.jcr</groupId>
+            <artifactId>jcr</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <!-- dependency>
```
hchiorean: this should be removed, since it's re-defined a few lines down

modeshape-jcr/src/main/resources/org/modeshape/jcr/repository-config-schema.json
```diff
@@ -28,6 +28,11 @@
             "default" : "10000",
             "description" : "The number of milliseconds to wait when a lock cannot be obtained on a node. In highly concurrent cases, this may be adjusted. Defaults to 10 seconds"
         },
+        "shellInterpreter" : {
```
hchiorean: maybe "enableShell" would be a better name ? Or something else which suggests a boolean value

hchiorean:
in addition to the above comments, this requires a few more changes:

- rebase on latest master and use 5.1-SNAPSHOT
- add `modeshape-shell` to the actual build (I don't see it among the list of modules)
- add the `modeshape-shell` module to the binary assembly descriptor
- remember that any configuration options have to be added to the WF kit as well (this doesn't need to be done as part of this PR necessarily, but it's something which has to be taken into account)

okulikov:
@hchiorean , thanks for comments. Will take it into account, it still a bit in debug mode.

okulikov:
@hchiorean , @rhauch  small question about WF-kit. Should we include `client` part into the kit or keep it separated? If yes, running start scripts should go to `bin` folder?

hchiorean:
@okulikov: would the server-side module be part of the WF kit ? If yes, what would that look like ? If we can include the server-side part of the console as a private or internal WF module (I think you can make a module private or internal via some sort of attribute in the descriptor) then we can add the client part as well.

And if we do that, yes, the scripts should probably go in the Wildfly `bin` folder.

okulikov:
@hchiorean , I think that would be good to include it. The another deal is that immediately after deployment shell MBean does not exists yet and not registered. It ups when repository actually starting. So it is impossible to access repository from shell client and this makes it useless.

rhauch:
@okulikov, this may be a bit of a hack, but could the client first use a REST call to check/poll whether the repository is running? If so, wouldn't that start the repository and then the shell client could connect??

okulikov:
@rhauch , like an option. I thought also about web-app for shell server side. Like http tunel

hchiorean:
@okulikov, @rhauch: can't all this be done using the `ModeShapeEngine` which is available in JNDI ?
In other words, for WF we could use an extension of the client (i.e. WF client) which uses the `Engine`
to:
a) check if the repository has been started and if not
b) starts the repository

Having a WF client would fit the pattern of the othe webapps for which we always have a WF "extension"

hchiorean:
@okulikov , @rhauch: using the REST service is a good idea, but the problem with that is it can easily be disabled by not configuring (including) the `modeshape-rest` webapp.

What about having a ModeShapeEngine generic MBean which can actually start a repository ? i.e. provide sort of a JMX interface for the ModeShapeEngine, not just each individual repository

okulikov:
@rhauch , @hchiorean Generic MBean looks good. I would say cooperation generic and individual MBeans just for isolation repository related communications

hchiorean:
my initial thought is that this EngineMBean could offer some basic functionality similar to what the engine does atm. For example: list all configured repositories and start a repository (if not already started).

The shell client could then use (initially) this bean to start/obtain a running repository.

@thoughts ?

okulikov:
agree

okulikov:
@hchiorean , @rhauch PR is updated. In this update:
- the first thing is that shell is started (optionally) with the `ModeshapeEngine` what gives access for configurations and for repositories which are not running yet; 
- 'client console' is wrapped with SSH server and it is accessible via standard SSH client so there is no 'client' module in this PR and user don't need any special client;
- SSH server implementation support only simple password authentication, user/password are configuration allows to define one user but guess it is enough for now;
- added implementation for commands related to the configuration task (commands are dynamic with accordance to the configuration schema) .

hchiorean:
maybe I missed something, but how did we decide to embed a full-blown SSH server inside `ModeShapeEngine` ? I remember discussing that we need to secure the JMX access, but surely there's got to be an easier way to do this.

This entire change seems way too complicated to me to be worth considering.

okulikov:
@hchiorean , in fact it simplifies a lot because JMX even for configuration in 'shell style' needs a session. And since we decided that it doesn't make sense for WF I just wrap console with SSH (SSH server is once class wich assigns startard i/o to console). It solves all visible problems and definitely allows to stop expense time for this issue.

hchiorean:
closing without merging, for the reason [described here](https://issues.jboss.org/browse/MODE-1660?focusedCommentId=13249728&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13249728)

PR #1539, opened 2016-03-22
MODE-2583 Fixes the handling of index definitions so that only index definitions present in the repository configuration are enabled and used

hchiorean:


rhauch:
Looks good.

PR #1540, opened 2016-03-23
MODE-2574, MODE-2575 Adds configuration support for the Cassandra and MongoDB binary stores

hchiorean:
The support is added both in the JSON and the Wildfly kit, based on the current code (i.e. in terms of configuration options). This also fixes a number of other JBoss AS kit issues, exposed by the enhancement of the JBoss AS subsystem tests.


rhauch:
Looks good.

PR #1541, opened 2016-03-28
MODE-2587 Updates backup and restore to perform the two operations in batches

hchiorean:
This should fix any memory issues caused by the previous implementation, where everything was loaded and written to in one big batch.

### DIFF
--- a/modeshape-jcr-api/src/main/java/org/modeshape/jcr/api/BackupOptions.java
+++ b/modeshape-jcr-api/src/main/java/org/modeshape/jcr/api/BackupOptions.java
@@ -27,6 +27,11 @@ public abstract class BackupOptions {
      * So, if each node requires about 200 bytes (compressed), the resulting files will be about 19 MB in size.
      */
     public static final long DEFAULT_DOCUMENTS_PER_FILE = 100000L;
+
+    /**
+     * The number of documents read form the persistent store and written in a backup file in one unit
+     */
+    public static final int DEFAULT_BATCH_SIZE = 10000;
     
     /**
      * Default backup options which will be used when a backup is performed without an explicit set of options.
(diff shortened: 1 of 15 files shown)


rhauch:
@hchiorean: Will there also be fixes applied to 4.x?

hchiorean:
The original [community issue](https://developer.jboss.org/message/953191) was raised around restore causing OOM issues (backup seemed to have worked). So users migrating to Modeshape 5 will benefit of the new restore code.

Migrating this to 4.x would not only mean a new 4.x release, but also different code (possibly) because the 4.x persistence (keys & documents retrieval and persistence) involve ISPN and slightly different API.

I suggest getting this in only for 5 atm. and if lots in users in the community have issues backing up a 4.x repository because of memory issues, then we can look at this for that branch.

rhauch:
@hchiorean, that sounds perfectly reasonable. Looks good.

awoods:
As a note, folks in the Fedora community are trying to upgrade from ModeShape 4 to 5 and are hitting this issue during `backup`. Without this patch being ported to 4.x, some in the Fedora community are unable move forward.
See: https://groups.google.com/d/msg/fedora-tech/_Yl1-h6lKpU/ytdoeFvQEwAJ

barmintor:
Especially given that backup/restore is the mechanism we have been proposing for migrating from 4.x to 5, a backport would be extremely useful.

scossu:
Echoing @awoods and @barmintor 's comment—we have a 4.x repository in production that we are unable to back up.

hchiorean:
note that this patch is not a straightforward port to 4.x. Because 4.x uses Infinispan, it most likely requires different code (which is semantically the same). 

If someone is willing to contribute this, we're happy to review and integrate it and try to do a `4.6.2` micro release (Maven only)

awoods:
To be clear, ModeShape recommends the use of `backup` and `restore` to migrate from ModeShape 4.x to 5.x, correct?
https://docs.jboss.org/author/display/MODE50/Migrating+from+3.x+and+4.x

Is there another option for those who can not use this approach due to the topic of this PR?

hchiorean:
Yes, backup & restore is the recommended way to migrate. I've re-opened [MODE-2587](https://issues.jboss.org/browse/MODE-2587) and targeted it for `4.6.2`.

However, since I'm tied up with smth. else atm., it will be a while before I can get back to looking at this. If in the meantime anyone wants to contribute, that would probably be the fastest way to get this ported.

Since this only affects `backup` (and not `restore`) most of the `backup` related changes from this PR can be backported as-is. The key difference is around [this line](https://github.com/ModeShape/modeshape/blob/4.x/modeshape-jcr/src/main/java/org/modeshape/jcr/BackupService.java#L428): after getting all the keys from ISPN, instead of loading all the documents into memory in the same method, a separate method should be created (similar to one from this PR) which loads only a subset of documents into memory.
Being a separate method, after each invocation, the GC could reclaim those objects since the method handler will be popped off the invocation stack.

awoods:
See: https://github.com/ModeShape/modeshape/pull/1604

PR #1545, opened 2016-04-21
MODE-2591 Changes the name of the datasource JNDI attribute for the DB persistence

hchiorean:
This should now match the other DS JNDI names used by the database binary store and JDBC metadata connector.


rhauch:
@hchiorean: Please confirm, but it appears that this will break any existing 5.0 configurations that use this. I understand the desire to correct this and eliminate potential problems for users (thus MODE-2591), but IMO any change like this needs to be backward compatible. IOW, it's okay to *add* the new property and even deprecate the old properties, but the subsystem still needs to be able to read and use the old properties.

hchiorean:
@rhauch: the change is intentional; deprecating the old property requires a lot of leftover boilerplate code to the point where I'd really not make this change at all (this was my original approach). 
Not only that, but there is no easy of elegant way to deprecate attributes in XML. We'd essentially be supporting 2 different names, with minor variation, for the same attribute.

My plan was to change the name altogether and update the release notes and documentation (it's a rather small change after all).

If you're not ok with it, I'm fine with moving/targeting this to the next major ModeShape release as opposed to deprecating the attributes, for the reasons mentioned above.

rhauch:
@hchiorean, it's up to you.

hchiorean:
@rhauch so you're ok with either option ?

rhauch:
@hchiorean, Sure. I just wanted to make sure the change was intentional.

PR #1546, opened 2016-04-26
MODE-2521 Updates Apache Tika to 1.12 and Apache POI to 3.13 (same version as the one used by Tika)

hchiorean:


rhauch:
Looks good.

PR #1547, opened 2016-04-28
MODE-2589 Fixes the handling of empty binaries by the DatabaseBinaryStore in the case of Oracle

hchiorean:


rhauch:
Looks good.

PR #1551, opened 2016-05-09
MODE-2596 Fixes the concurrent initialization of the same workspace cache with user transactions

hchiorean:
The problem in this particular case was the fact that a user transaction off a certain thread might not have been committed while  another thread was seeing the ws cache as already initialized. 

The solution was to make sure ws cache initialization takes place inside local transactions by suspending/resuming any potential outside transactions.

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/query/engine/process/BufferingSequence.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/query/engine/process/BufferingSequence.java
@@ -118,24 +118,7 @@ public abstract class BufferingSequence extends DelegatingSequence {
         Batch batch = sequence.nextBatch();
         int batchSize = 0;
         Object value = null;
-        while (batch != null && batchSize == 0) {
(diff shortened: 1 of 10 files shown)


rhauch:
Looks good.

modeshape-jcr/src/main/java/org/modeshape/jcr/query/engine/process/BufferingSequence.java
```diff
@@ -118,24 +118,7 @@ protected int loadAll( NodeSequence sequence,
...
-                } else if (value != null) {
-                    buffer.put(value, createRow(batch));
-                } else if (rowsWithNullKey != null) {
-                    rowsWithNullKey.addIfAbsent(createRow(batch));
-                }
-                ++batchSize;
-            }
-            batch = sequence.nextBatch();
-        }
+        boolean firstBatchCounted = false;
```
Waize: See https://issues.jboss.org/browse/MODE-2677

PR #1553, opened 2016-05-11
MODE-2593 Adds datasource support to the Database Binary Store

hchiorean:
This uses the same datasource implementation (Hikari) that is being used by the relational persistent storage. This artifact has also been promoted as a top-level artifact in the WF kit.


rhauch:
Looks good.

PR #1502, opened 2016-01-05
MODE-2552 Adds sequencer for several audio formats

RobSis:
Initial implementation of audio sequencer based on jAudioTagger.

PR #1554, opened 2016-05-13
MODE-2553 Adds sequencer for several video formats

RobSis:
New pull request against the master branch. 

First review was done here:
https://github.com/ModeShape/modeshape/pull/1503

PR #1555, opened 2016-05-13
MODE-2552 Adds sequencer for several audio formats

RobSis:
New pull request against the master branch. 

First review was done here:
https://github.com/ModeShape/modeshape/pull/1502


hchiorean:
looks good
@rhauch  ?

rhauch:
Looks good to me.

hchiorean:
@RobSis: this no longer merges cleanly after merging https://github.com/ModeShape/modeshape/pull/1554
Please fix it in your local branch and force update (`push -f`) this PR. Thanks.

hchiorean:
@rhauch: bringing this in, will present a bit of a conundrum: we already have an `mp3` sequencer which uses the same library and does some (modest) mp3 parsing. 

This sequencer supports multiple audio formats (mp3 included) but with a different (more detailed and generic) structure.

So IMO we have a couple of options:
1.add this and keep both sequencers side-by-side while deprecating the `mp3` sequencer (marking its classes as deprecated)
2.merge the 2 sequencers into one (named `audio` which is more generic)

(1) means no changes are required as this one will be a new addition
(2) means if anyone (*if*) is using the mp3 sequencer they'd have to change their configuration

wdyt ?

RobSis:
Note that I'm adding this to binary distributions again. 

I've noticed you removed video sequencer from it, so I'm not sure whether this sequencer should be there...

hchiorean:
this should be part of the binary dist (as is the case with the other sequencers as well).

The problem I have with the video sequencer (as described [here](https://issues.jboss.org/browse/MODE-2553?focusedCommentId=13238390&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13238390)) is that the size of its `io.humble` dependency is about `115 MB`. This is *way way* too big for a sequencer for us to bundle and deliver.

hchiorean:
@rhauch any thoughts on https://github.com/ModeShape/modeshape/pull/1555#issuecomment-219670791 ?

hchiorean:
@RobSis: with the merging of the other sequencers, this again needs to be brought up to date with `master`.

PR #1556, opened 2016-05-16
MODE-2535: Readonly user is not allowed in Web Explorer

okulikov:
This PR handles case when user has no permission to read access control content. In this case access list is not displayed rather then exception is thrown.


hchiorean:
looks good

PR #1557, opened 2016-05-17
MODE-2585, MODE-2586 Updates Lucene to 6.0.Final and changes some of the Lucene querying code

hchiorean:
The following are contained in this commit:

a) changing the version to 6.0.Final and making it a private dependency of the `lucene-index-provider` in WF means we should now be independent of whatever version of Lucene WF/EAP comes with. ModeShape will no longer install Lucene as a first-class dependency in WF.

b) with the above change, the ES index provider will use its own separate version (in this case 5.3.1.Final) and from now on can be evolved independently of the `lucene-index-provider`

c) changing the design around the `Filter#Results` batching mechanism so that index providers can opt for truly lazy index results when SKIP is used. In the case of Lucene this means that the Lucene query is only run once after the first N batches have been skipped.

d) changing the logic of the Lucene query making use of `BitSets` and bulk-loading documents at the end of the scoring phase as opposed to loading each document during the collection phase.


deploy/jbossas/kit/jboss-wf/org/modeshape/index-provider/lucene/main/module.xml
```diff
@@ -18,10 +18,15 @@
 <module xmlns="urn:jboss:module:1.3" name="org.modeshape.index-provider.lucene">
     <resources>
         <resource-root path="modeshape-lucene-index-provider-${project.version}.jar" />
+        <resource-root path="lucene-analyzers-common-${version.org.apache.lucene}.jar" />
+        <resource-root path="lucene-core-${version.org.apache.lucene}.jar" />
+        <resource-root path="lucene-queries-${version.org.apache.lucene}.jar" />
+        <resource-root path="lucene-queryparser-${version.org.apache.lucene}.jar" />
+        <resource-root path="lucene-sandbox-${version.org.apache.lucene}.jar" />
+        <resource-root path="lucene-backward-codecs-${version.org.apache.lucene}.jar" />
```
rhauch: I understand why we'd want to remove the higher-level dependency on Lucene and instead have that dependency restricted to the Lucene index provider, but why are we _embedding_ the Lucene libraries inside the provider module rather than simply having the provider module depend on a separate Lucene module?

deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/AddIndexProvider.java
```diff
@@ -40,7 +41,13 @@
 
     public static final AddIndexProvider INSTANCE = new AddIndexProvider();
 
-    private static final Logger LOG = Logger.getLogger(AddIndexProvider.class.getPackage().getName());
+    private static final Map<String, String> MODULE_NAMES_BY_PROVIDER_ALIAS = new HashMap<String, String>() {
+        {
+            put("org.modeshape.jcr.index.lucene.LuceneIndexProvider", "org.modeshape.index-provider.lucene");
+            put("org.modeshape.jcr.index.elasticsearch.EsIndexProvider", "org.modeshape.index-provider.elasticsearch");
+            put("org.modeshape.jcr.index.local.LocalIndexProvider", "org.modeshape");
```
rhauch: Wouldn't it be useful to _also_ add aliases such as `lucene`, `elasticsearch`, and `local`? Seems like those simple aliases would be easier for people to use correctly.

deploy/jbossas/kit/jboss-wf/org/modeshape/index-provider/lucene/main/module.xml
hchiorean: IMO it only make sense to have a top-level Lucene libary in WF if more than one module depends on it (or parts of it). Note that WF already has a `org.apache.lucene` module but with a different version.

This type of changes makes clear the intent that this is a private dependency of a module, regardless of what's in WF.

The other reason for this change is that in the case of EAP there's very strict monitoring of the modules which are installed/deployed. Having a top-level `org.apache.lucene` module with a different version than the default one (which is built from sources) will probably create a lot of productization issues.

rhauch: Ok.

deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/AddIndexProvider.java
hchiorean: this line `String fqProviderClass = RepositoryConfiguration.getBuiltInIndexProviderClassName(providerClassName);` (see below) already accounts for the shorthand aliases. 

The above change is only to make sure users aren't forced to always have the `module` attribute set by hand.

rhauch: Ok.

hchiorean:
@rhauch: are you ok with merging this ?

rhauch:
@hchiorean, yes, this looks good.

PR #1559, opened 2016-05-17
MODE-2601 enable configurable node handling by FileSystemConnector subclasses

magibney:
addresses [MODE-2601](https://issues.jboss.org/browse/MODE-2601)

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/connector/filesystem/FileSystemConnector.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/connector/filesystem/FileSystemConnector.java
@@ -481,16 +481,31 @@ public class FileSystemConnector extends WritableConnector implements Pageable {
         return fileFor(id).exists();
     }
 
+    /**
+     * Hook for processing other node types (not resources or regular files)
+     * @param id
+     * @param file
+     * @param isRoot
+     * @param isResource
+     * @return DocumentWriter if node is recognized and processed; otherwise
+     * null (signaling subsequent fallback to default processing)
+     */
+    protected DocumentWriter writerFor(String id, File file, boolean isRoot, boolean isResource) {
+        return null;
+    }
+
     @Override
     public Document getDocumentById( String id ) {
         File file = fileFor(id);
         boolean isRoot = isRoot(id);
         // make sure the root folder is never excluded....
-        if (!isRoot && (isExcluded(file) || !file.exists())) return null;
+        if (!isRoot && (isExcluded(file) || !acceptFile(file))) return null;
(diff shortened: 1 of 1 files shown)


modeshape-jcr/src/main/java/org/modeshape/connector/filesystem/FileSystemConnector.java
```diff
@@ -481,16 +481,30 @@ public boolean hasDocument( String id ) {
...
 
+    /**
+     * Hook for processing other node types (not resources or regular files)
+     * @param id
+     * @param file
+     * @param isRoot
+     * @return true if node is recognized and processed; false causes fallback
+     * to processing as a directory.
+     */
+    protected DocumentWriter otherType(String id, File file, boolean isRoot, boolean isResource) {
```
hchiorean: what about renaming this to `writerFor(...)` ? Seems like a better name given the intent of the method

modeshape-jcr/src/main/java/org/modeshape/connector/filesystem/FileSystemConnector.java
```diff
@@ -481,16 +481,30 @@ public boolean hasDocument( String id ) {
...
+
     @Override
     public Document getDocumentById( String id ) {
         File file = fileFor(id);
         boolean isRoot = isRoot(id);
         // make sure the root folder is never excluded....
-        if (!isRoot && (isExcluded(file) || !file.exists())) return null;
+        if (!isRoot && (isExcluded(file) || !acceptFile(file))) return null;
         boolean isResource = isContentNode(id);
         DocumentWriter writer = null;
```
hchiorean: with the above methods, this could be changed to:
`writer = writerFor(...)` and then the check would be simply `if (writer != null)` or else do the regular logic

hchiorean:
@magibney: this looks like a good change (I only had some minor comments)

magibney:
@hchiorean: thanks for your response and suggestions! I've pushed a commit changing the method name as you suggested, fixing some documentation, and reordering/separating out some statements for clarity.

hchiorean:
thanks, this looks good now. 
Can you please squash the 2 commits and force update the PR (`push -f`) ? thanks

rhauch:
This will be great once the commits are squashed.

magibney:
@hchiorean: ok squashed now, thanks! In addition to the previously-reviewed changes, the squashed commit also moves (and minimally modifies) the MODE-1768 comment to keep it next to the default logic to which it refers (now resides in the new `acceptFile` method, ~line 600).

PR #1560, opened 2016-05-19
MODE-2551 Adds EPUB 3.0 sequencer

RobSis:
New pull request against the master branch. 

First review was done here:
https://github.com/ModeShape/modeshape/pull/1501


sequencers/modeshape-sequencer-epub/src/main/java/org/modeshape/sequencer/epub/EpubMetadata.java
```diff
@@ -0,0 +1,346 @@
...
+     * The sequencer first examines the container to get the list of all rootfiles.
+     * Then it process each rootfile and extracts the metadata.
+     */
+    public boolean check() throws Exception {
+        // we need to create a copy of the file, because the container file
+        // might be at the end of the stream.
+        File fileCopy = File.createTempFile("modeshape-sequencer-epub", ".tmp");
+        IoUtil.write(in, new BufferedOutputStream(new FileOutputStream(fileCopy)));
+
+        ZipInputStream zipStream = new ZipInputStream(new FileInputStream(fileCopy));
```
hchiorean: how about using a `try-with-resources` idiom ? Seems more robust

sequencers/modeshape-sequencer-epub/src/main/java/org/modeshape/sequencer/epub/EpubMetadata.java
```diff
@@ -0,0 +1,346 @@
...
+        List<String> rootfiles = getRootfiles(zipStream);
+        zipStream.close();
+
+        zipStream = new ZipInputStream(new FileInputStream(fileCopy));
```
hchiorean: same comment as above; in addition, perhaps the construction of the `zipStream` can be deferred to only when `!rootFiles.isEmpty()`. It seems it's not required in any other case.

sequencers/modeshape-sequencer-epub/src/main/java/org/modeshape/sequencer/epub/EpubMetadata.java
```diff
@@ -0,0 +1,346 @@
...
+        return rootfiles;
+    }
+
+    /**
+     * Read the content of the ZipEntry without closing the stream.
+     */
+    private ByteArrayOutputStream getZipEntryContent(
+            ZipInputStream zipStream,
+            ZipEntry entry ) throws IOException {
+        ByteArrayOutputStream content = new ByteArrayOutputStream();
```
hchiorean: `try-with-resources` would probably be cleaner

sequencers/modeshape-sequencer-epub/src/main/java/org/modeshape/sequencer/epub/EpubMetadata.java
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    public List<EpubMetadataProperty> getAlternateScript() {
+        return alternateScript;
+    }
+
+    /**
+     * Simple NamespaceContext which can resolve single namespace.
+     */
+    private class SimpleNSC implements NamespaceContext {
```
hchiorean: there's already a `org.modeshape.common.xml.SimpleNamespaceContext` in `modeshape-common`. Perhaps we can use that here instead of having this inner class.

sequencers/modeshape-sequencer-epub/src/main/java/org/modeshape/sequencer/epub/EpubMetadataSequencer.java
```diff
@@ -0,0 +1,198 @@
...
+                        values.add(vf.createValue((String) val));
+                    }
+                }
+                if (values.size() > 0) {
+                    Value[] valuesArray = new Value[values.size()];
+                    values.toArray(valuesArray);
+                    node.setProperty(propertyName, valuesArray);
+                }
+            } else {
+                throw new IllegalArgumentException(String.format("The value of the property %s has unknown type and couldn't be saved.", propertyName));
```
hchiorean: do we really want to throw an exception here ? Wouldn't perhaps logging a warning and continuing be more appropriate ?

sequencers/modeshape-sequencer-epub/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,20 @@
...
+log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
+log4j.appender.stdout.layout.ConversionPattern=%d{ABSOLUTE} %5p %m%n
+
+# Root logger option
+log4j.rootLogger=INFO, stdout
+
+# Set up the default logging to be INFO level, then override specific units
+log4j.logger.org.modeshape=INFO
+
+# Turn off the serialization protocol used in Hibernate Search (for clustering)
```
hchiorean: the logging statements below this comment can be removed; ModeShape 5 doesn't use HSearch or ISPN anymore.

hchiorean:
with the exception of the comments, this looks good. Nice work @RobSis

sequencers/modeshape-sequencer-epub/src/main/java/org/modeshape/sequencer/epub/EpubMetadataSequencer.java
RobSis: This is done the same way as all my already merged sequencers...
Maybe I could leave it like this for now and make a new PR for changing it everywhere at once?

The same goes for the log4j.properties file.

hchiorean: ok, that's fine.

I have to add most of the other sequencers to the WF kit anyway, so I can make this change myself. Just have a look at the other comments please.

sequencers/modeshape-sequencer-epub/src/main/java/org/modeshape/sequencer/epub/EpubMetadata.java
```diff
@@ -0,0 +1,317 @@
...
+        // we need to create a copy of the file, because the container file
+        // might be at the end of the stream.
+        File fileCopy = File.createTempFile("modeshape-sequencer-epub", ".tmp");
+        IoUtil.write(in, new BufferedOutputStream(new FileOutputStream(fileCopy)));
+
+        List<String> rootfiles = new ArrayList<>();
+        try (ZipInputStream zipStream =
+                new ZipInputStream(new FileInputStream(fileCopy))) {
+            rootfiles = getRootfiles(zipStream);
+            zipStream.close();
```
hchiorean: this is not required; that's the whole purpose of `try-with-resources`

sequencers/modeshape-sequencer-epub/src/main/java/org/modeshape/sequencer/epub/EpubMetadata.java
```diff
@@ -0,0 +1,317 @@
...
+            ZipInputStream zipStream,
+            ZipEntry entry ) throws IOException {
+        try (ByteArrayOutputStream content =
+                new ByteArrayOutputStream()) {
+            byte[] bytes = new byte[(int) entry.getSize()];
+            int read;
+            while ((read = zipStream.read(bytes, 0, bytes.length)) != -1) {
+                content.write(bytes, 0, read);
+            }
+            content.flush();
```
hchiorean: this is not required; the `ByteArrayOutputStream#flush` method is empty anyway. Writing to a `ByteArrayOutputStream ` is done immediately

hchiorean:
looks good
@rhauch  ?

sequencers/modeshape-sequencer-epub/src/main/java/org/modeshape/sequencer/epub/EpubMetadataSequencer.java
```diff
@@ -0,0 +1,198 @@
...
+                ValueFactory vf = node.getSession().getValueFactory();
+                List<Value> values = new ArrayList<Value>();
+                for (Object val : (List<?>) value) {
+                    if (val instanceof String) {
+                        values.add(vf.createValue((String) val));
+                    }
+                }
+                if (values.size() > 0) {
+                    Value[] valuesArray = new Value[values.size()];
+                    values.toArray(valuesArray);
```
hchiorean: actually I missed this: you're ignoring the result of the `toArray` method which is what you really want.
So the correct code would be `node.setProperty(propertyName, values.toArray(new Value[values.size()]))`

RobSis: Thank you. 

Note that this is again the case in all previously merged PRs...

hchiorean:
@rhauch ?

sequencers/modeshape-sequencer-epub/src/main/java/org/modeshape/sequencer/epub/EpubMetadataSequencer.java
```diff
@@ -0,0 +1,196 @@
...
+                addEpubMetadataProperties(sequencedNode, EpubMetadataLexicon.LANGUAGE, metadata.getLanguage());
+                addEpubMetadataProperties(sequencedNode, EpubMetadataLexicon.IDENTIFIER, metadata.getIdentifier());
+                addEpubMetadataProperties(sequencedNode, EpubMetadataLexicon.DESCRIPTION, metadata.getDescription());
+                addEpubMetadataProperties(sequencedNode, EpubMetadataLexicon.PUBLISHER, metadata.getPublisher());
+                addEpubMetadataProperties(sequencedNode, EpubMetadataLexicon.DATE, metadata.getDate());
+                addEpubMetadataProperties(sequencedNode, EpubMetadataLexicon.RIGHTS, metadata.getRights());
+                return true;
+            }
+        } catch (Exception e) {
+            getLogger().debug(e, "Couldn't process stream.");
```
rhauch: Should this be debug-level or error-level?

RobSis: Should be error level...

PR #1564, opened 2016-05-27
MODE-2602 Updates the WF kit so that module names for built-in components don't require explicit configuration

hchiorean:


rhauch:
@hchiorean, looks good.

PR #1566, opened 2016-06-01
MODE-2606 Updates the versions of some common ModeShape EAP7/WF 10 components

hchiorean:


boms/modeshape-bom-embedded/pom.xml
```diff
@@ -493,164 +493,6 @@
...
-                    <exclusion>
-                        <artifactId>hornetq-core</artifactId>
-                        <groupId>org.hornetq</groupId>
-                    </exclusion>
-                    <exclusion>
-                        <artifactId>hornetq-commons</artifactId>
-                        <groupId>org.hornetq</groupId>
-                    </exclusion>
-                </exclusions>
             </dependency>
```
rhauch: Are these all finally excluded now in the artifact?

rhauch:
One question above, but otherwise looks good.

boms/modeshape-bom-embedded/pom.xml
hchiorean: yup, as far as I can tell `org.jboss.narayana.jta:narayana-jta` doesn't have any additional compile-time transitive dependencies.

PR #1567, opened 2016-06-03
MODE-2604 Adds support for unnamed CHECK constraints to the DDL sequencer

hchiorean:


sequencers/modeshape-sequencer-ddl/src/test/resources/ddl/create_table_check_constraint.ddl
```diff
@@ -0,0 +1,21 @@
+--  SQL Server / Oracle / MS Access, not named
+--  Note that starting with 5.7 MySQL does not support CHECK constraints
```
rhauch: I'm not sure this comment is correct. MySQL 5.5-5.7 [all support](http://dev.mysql.com/doc/refman/5.7/en/create-table.html) CHECK constraints, though whether they are _used_ depends on the storage engine.

rhauch: Well, it does not appear that MySQL 5.7 does not support _named_ CHECK constraints, as in the `PersonsNamed` example below.

hchiorean: well, the MySQL 5.7 documentation states:
> The CHECK clause is parsed but ignored by all storage engines. 

As such, I don't think there's anything MySQL specific the parser should do. i.e. the default `CHECK` handling (which seems to be the same for the other supported DBs) should be enough

hchiorean:
@rhauch: if you prefer, I can remove that particular comment from the test file

sequencers/modeshape-sequencer-ddl/src/test/resources/ddl/create_table_check_constraint.ddl
rhauch: I agree. But the comment is not correct and should be fixed or removed. I suggest removing it.

rhauch:
@hchiorean once that's removed, it looks good to me.

hchiorean:
amended the commit and merged in `master`

PR #1568, opened 2016-06-07
MODE-2607 Fixes incorrect behaviour when using user transactions

hchiorean:
The bug was caused by the fact that the transactional ws cache was being cleared after each `session.save()` call, even if  the user transaction might still not have been committed. This meant that subsequent `session.save()` calls could get/see transient changes as the "latest persisted information".


rhauch:
Looks great.

PR #1569, opened 2016-06-10
MODE-2094: Access list support fot CMIS connector

okulikov:
This PR allows to read access lists defined by CMIS repository.

### DIFF
--- a/connectors/modeshape-connector-cmis/src/test/java/org/modeshape/connector/cmis/CmisConnectorIT.java
+++ b/connectors/modeshape-connector-cmis/src/test/java/org/modeshape/connector/cmis/CmisConnectorIT.java
@@ -236,7 +236,7 @@ public class CmisConnectorIT extends MultiUseAbstractTest {
         assertTrue(date != null);
     }
 
-//     @Test
+    @Test
     public void shouldAccessModificationDatePropertyForFolder() throws Exception {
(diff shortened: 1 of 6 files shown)


connectors/modeshape-connector-cmis/src/test/java/org/modeshape/connector/cmis/CmisConnectorIT.java
```diff
@@ -234,7 +236,7 @@ public void shouldAccessCreationDatePropertyForFolder() throws Exception {
         assertTrue(date != null);
     }
 
-    // @Test
+//     @Test
```
hchiorean: should this be commented out ?

hchiorean:
@okulikov: with the exception of the question, this looks good to me

okulikov:
@hchiorean , it was commented and actually I don't remember why. When I uncomment this test fails but not due to access list. Do we want to fix this bug in this PR as well?

hchiorean:
sure. If there's a bug somewhere in the connector, we should fix it. You can add a separate commit to this PR for the fix.

okulikov:
@hchiorean , do we want to squash both commits?

hchiorean:
@okulikov: no, I think a separate commit is fine as it underlines that there's a fix for a bug
@rhauch: this looks good to me, are you ok with merging ?

PR #1570, opened 2016-06-13
MODE-2610 Fixes another transactional issue

hchiorean:
The bug was caused by the fact that not all the nodes were refreshed from the document store (i.e. the latest persisted version) on successive `session.save()` calls made from the same ongoing transaction


rhauch:
Looks good.

PR #1571, opened 2016-06-14
MODE-2609 Fixes the type of the jcr:childVersionHistory property

hchiorean:
The property should be a REFERENCE, not a STRING. The code was changed so that if the value for the property was stored as a STRING, the node to which it points can still be read back.


rhauch:
Looks good. 

Please ignore the CI failure above. The Travis CI build is not yet configured, so it's always failing and I'm currently working on completing that work.

PR #1572, opened 2016-06-14
MODE-2599 Added Travis-CI build

rhauch:
The build just runs `mvn -s settings.xml clean install -Pintegration`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+language: java
+
+# Travis by defaults runs the following `install` command:
+#
+#    install: mvn isntall -DskipTests=true -Dmaven.javadoc.skip=true -B -V
+#
+# though we have to add `-s settings.xml` for our build to work:
+#
+#    install: mvn -s settings.xml install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
+#
+# But we actually want to skip this so that it doesn't take as long to fail
+# so we skip the `install` phase:
+install: true
+
+# We have to use our local `settings.xml`, and the clustering tests require IPv4 on the Travis infra
+script: mvn -s settings.xml clean install -Pintegration -DpreferIpv4
+
+jdk:
+  - oraclejdk8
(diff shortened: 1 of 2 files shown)


hchiorean:
+1 (except that build job seems to be failing)

hchiorean:
@rhauch: the missing artifacts should (hopefully) be resolvable by updating `settings.xml` and adding & enabling the following repo:
```
<profile>
            <id>jboss-thirdpartyuploads-repository</id>
            <repositories>
                <repository>
                    <!--This repository is required for the Fungal artifact used by ModeShape JCA -->
                    <id>jboss-thirdpartyuploads-repository</id>
                    <name>JBoss Third Party Uploads Repository</name>
                    <url>http://repository.jboss.org/nexus/content/repositories/thirdparty-uploads/</url>
                    <layout>default</layout>
                    <releases>
                        <enabled>true</enabled>
                        <updatePolicy>never</updatePolicy>
                    </releases>
                    <snapshots>
                        <enabled>false</enabled>
                        <updatePolicy>never</updatePolicy>
                    </snapshots>
                </repository>
            </repositories>
        </profile>
```

which is required for the JCA stuff anyway

rhauch:
@hchiorean, yes, the problem is that Travis is running a pre-build `mvn` command to install dependencies, and I forgot to override that. Trying again.

rhauch:
The latest problem appears to be that the Git connector tests are assuming the existence of a local git repo that has tags, but the Travis build does not include tags when it clones the repo prior to the build using:

    git clone --depth=50 https://github.com/ModeShape/modeshape.git ModeShape/modeshape

IMO, if the tests need a local copy of the Git repo, then they should clone the Git repo into a target directory for use by the tests.

hchiorean:
Merged in `master`, although the current failure is 
`The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).` and is due to Maven logging a lot of output 

There are several Travis issues on this topic, including an open feature request to allow making this limit configurable, but there is no clear solution yet.

PR #1576, opened 2016-06-23
MODE-2513: Add pager for node list in Web-explorer

okulikov:
This PR adds pager for the list of child nodes


hchiorean:
looks good to me

PR #1577, opened 2016-06-28
MODE-2510 Fixes various issues around SNS reordering, including keeping [jcr:path] indexes up-to-date in such cases

hchiorean:


rhauch:
Looks good.

PR #1578, opened 2016-07-01
MODE-2615 Fixes the BsonDataInput to handle multi-byte UTF-8 characters

amccormic:
This fix ensures that the `limit` is set correctly when there is already data in the `ByteBuffer` from a previous decode operation. This happens when a multi-byte UTF-8 character could not be consumed by the decoder and is left in the buffer for the next iteration.


hchiorean:
looks like the correct fix, thank @amccormic. I'll create a new (more compact) PR with a few more test cases shortly.

PR #1580, opened 2016-07-08
Mode 2613 : Modeshape persistence test

okulikov:
This PR for provides module skeleton:
- builds docker images for MySQL, Postgres and Oracle databases;
- runs "dummy" test multiple times against each database.

If this skeleton looks good then I will start to add concrete tests.


modeshape-persistence-tests/pom.xml
```diff
@@ -0,0 +1,332 @@
...
+                            <systemPropertyVariables>
+                            <!-- Make these available to the tests via system properties -->
+                                <database.driver>org.postgresql.Driver</database.driver>
+                                <database.url>jdbc:postgresql://localhost/${database.name}</database.url>
+                                <database.user>${database.user}</database.user>
+                                <database.password>${database.password}</database.password>
+                            </systemPropertyVariables>
+                        </configuration>
+                    </execution>
+                    <execution>
```
hchiorean: if the `verify` goal required ? In other words, aren't the above `integration-test` goals enough ?

hchiorean:
other than a minor comment, this looks good to me

modeshape-persistence-tests/pom.xml
rhauch: Yes, it is required. Failsafe always runs:

* `pre-integration-test` to set up the test infrastructure (which in our case is the Docker containers)
* `integration-test` to run the integration tests using Surefire;
* `post-integration-test` to always tear down the test infrastructure (which in our case is the Docker containers);
* `verify` to check whether any integration tests failed, and if so stops Maven.

This is how Failsafe ensures that the test infrastructure is always torn down before Maven is stopped. See the [Failsafe documentation](http://maven.apache.org/surefire/maven-failsafe-plugin/) for details.

modeshape-persistence-tests/pom.xml
```diff
@@ -0,0 +1,332 @@
...
+                        <goals>
+                            <goal>integration-test</goal>
+                        </goals>
+                        <configuration>
+                            <systemPropertyVariables>
+                            <!-- Make these available to the tests via system properties -->
+                                <database.driver>org.postgresql.Driver</database.driver>
+                                <database.url>jdbc:postgresql://localhost/${database.name}</database.url>
+                                <database.user>${database.user}</database.user>
+                                <database.password>${database.password}</database.password>
```
rhauch: Why do we need to define `database.user` and `database.password` here when they are defined exactly the same as the system properties defined in the `configuration` section? Seems like they can be removed.

modeshape-persistence-tests/pom.xml
```diff
@@ -0,0 +1,332 @@
...
+                    <execution>
+                        <id>integration-test-postgres</id>
+                        <goals>
+                            <goal>integration-test</goal>
+                        </goals>
+                        <configuration>
+                            <systemPropertyVariables>
+                            <!-- Make these available to the tests via system properties -->
+                                <database.driver>org.postgresql.Driver</database.driver>
+                                <database.url>jdbc:postgresql://localhost/${database.name}</database.url>
```
rhauch: Using `localhost` will only work on Linux, and will not currently work on OS X or Windows because on those platforms Docker is run in a virtual machine. Doesn't the Fabric8 Maven plugin set the `${docker.host}` on all platforms, even when Docker on Linux doesn't set `DOCKER_HOST` env variable? If so, then this should probably be:

    <database.url>jdbc:postgresql://${docker.host}/${database.name}</database.url>

modeshape-persistence-tests/pom.xml
```diff
@@ -0,0 +1,332 @@
...
+                    <execution>
+                        <id>integration-test-mysql</id>
+                        <goals>
+                            <goal>integration-test</goal>
+                        </goals>
+                        <configuration>
+                            <systemPropertyVariables>
+                            <!-- Make these available to the tests via system properties -->
+                                <database.driver>com.mysql.jdbc.Driver</database.driver>
+                                <database.url>jdbc:mysql://localhost:3306/${database.name}</database.url>
```
rhauch: Using `localhost` will only work on Linux, and will not currently work on OS X or Windows because on those platforms Docker is run in a virtual machine. Doesn't the Fabric8 Maven plugin set the `${docker.host}` on all platforms, even when Docker on Linux doesn't set `DOCKER_HOST` env variable? If so, then this should probably be:

    <database.url>jdbc:mysql://${docker.host}:${database.port}/${database.name}</database.url>

rhauch:
@okulikov, I think the properties need to be fixed; see my comments above.

Also, IIUC when this module is run by the Maven build, it will run all the integration tests against MySQL and then run them again using PostgreSQL. Is there a way to just run one or the other? Do we want this or need this? I guess it depends on whether we will want to debug a problem with just one of the DBMSes.

Perhaps I'm wrong, but I thought the only way to have some flexibility (if we want it) is to use profiles. You might have the default config for the Failsafe plugin run just against MySQL (so by default that's all that happens), a profile that just runs PostgreSQL, and another profile (e.g,. named `assembly` since our release uses that) that runs them all.

okulikov:
@rhauch , I agree regarding profiles.

okulikov:
@rhauch , @hchiorean , Do we want to run docker container related to profile (and all container for assembly profile) or it is good if all containers are started irrespective of profile?

hchiorean:
IMO this module should not be related to the `assembly` or `integration` profiles. It should do "its own thing"

modeshape-persistence-tests/pom.xml
```diff
@@ -0,0 +1,417 @@
...
+            <artifactId>junit</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-library</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+        </dependency>
```
hchiorean: why do we want the `mockito` dependency here ? IMO the tests should not be mocking anything

modeshape-persistence-tests/pom.xml
```diff
@@ -0,0 +1,417 @@
...
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
+        <profile>
+            <id>assembly</id>
```
hchiorean: IMO the assembly profile should not be here. Moreover, this module should not be run neither as part of `assembly` nor `disitribution`. This module should me more akin to `performance`

modeshape-persistence-tests/src/test/resources/data/simple.json
```diff
@@ -0,0 +1,244 @@
+{
```
hchiorean: is this json used anywhere ?

modeshape-persistence-tests/pom.xml
okulikov: We can rename it to `persistense` similar to 'permormance' but keep dedicated profiles here as well? 

Profiles give us some more flexibility

modeshape-persistence-tests/src/test/resources/data/simple.json
okulikov: not yet, I will clean up all not used things after completing test cases

modeshape-persistence-tests/pom.xml
hchiorean: I agree that it's fine to have profiles for the different RDBMS with some default (e.g. H2). If we want a profile to run all the tests against all supported DBs, let's rename it to 'all' or something along those lines

okulikov:
@hchiorean , @rhauch , Just few thoughts about test cases. I guess all of them should be against persistence, so each test case should also restart repository and check that all changes are still in place, right?

hchiorean:
most test cases should IMO drop the table `after` and create it `before` (both are JSON configuration settings). This way each persistence test runs against an empty repository

In the case of restart tests, we obviously want a different setting which does not drop the table on exit, so the first thing which comes to my mind is having a different test suite class (not unlike `JcrRepositoryStartupTest`)  and essentially controlling the `dropOnExit` behavior from the JSON config via a System property. 

This way each test suite can decide `@BeforeClass` which approach it wants.

rhauch:
Actually, we shouldn't expect the tests are executed in a particular order; that's generally bad practice and hard to work with in the IDE. Instead, I think if the test restarts the repo to verify state after restart, it should do it within the tast case method. Let's keep things simple.

rhauch:
IOW, have the `@Before` drop the table every time.

hchiorean:
>IOW, have the @Before drop the table every time. 

that won't really work: if you've configured a repository with `dropOnExit : true`, once you issue a `shutdown` command, it will drop your entire table (and data). Therefore it would be impossible to test for example that data is correctly persisted across restarts (e.g. write - shutdown - read scenarios)

Note that there's no way to issue a table drop explicitly in `@Before`, so the only option is to rely on repository behavior.

rhauch:
Would using JDBC not work to set up and tear down the databases when needed?

rhauch:
Drop statements are pretty straightforward. Even if not, a simple property file would work well enough.

hchiorean:
>Drop statements are pretty straightforward. Even if not, a simple property file would work well enough. 

it would work, but don't we want to test the entire repository behavior ? The `createOnStart` and `dropOnExit` settings are important IMO. So we could set the first one to `true` and the second one to `false` (which are the defaults IIRC), but don't we want to test for example that shutting down a repository with `dropOnExit=true` followed by a startup correctly drops and recreates the table on all RDBMS ?

I'm fine with having a separate test suite for the above cases and using JDBC directly in all other cases, but IMO the above scenarios are very important to test on all DBs. 

My point above was that having a JSON configuration for example with `dropOnExit=true` would impact all test cases which use this repo (and we don't want multiple JSON configuration files, which is why I was suggesting we configure it like `dropOnExit=${dropOnExit}` so that each suite can fine-tune this behavior.

rhauch:
Maybe I'm not being clear. If the `@Before` and `@After` methods simply cleaned up the content of the database, then each test method is entirely self-contained and would be responsible for starting the repository using a configuration of its own choosing, testing the repository, and then shutting it down. Perhaps helper methods make this a bit easier, but maybe not.

So some test methods might use a configuration with `createOnStart=true` and `dropOnExit=true`, and, after shutting down the repository, verify with JDBC that the database is indeed cleaned up. Other test methods might use a configuration with `createOnStart=true` and `dropOnExit=false`, and, after shutting down the repository, either verify with JDBC that the state of the database was maintained or start the repository again with `createOnStart=true` (or `false`, depending upon what needed to be tested) and verify the data the test method previously saved in the repository is still there.

So, yes, we should be using and testing the behavior of ModeShape (e.g., the `createOnStart` property), but we may want or need to use JDBC at various times to ensure the database state is what we expect it to be (including before and after a test is run). Sometimes we can infer what happened based upon ModeShape behavior (e.g., we found our data after shutting down and restarting, so obviously the table was not deleted upon shutdown), but other times we'll want to be more explicit.

rhauch:
BTW, make sure that tests are not run in parallel, because if they are they will likely stomp on each other's persistent storage.

okulikov:
@hchiorean , @rhauch , I did squash commits so here are already test cases for review. All of them are based on existing unit tests and adopted for testing against persistence.

modeshape-persistence-tests/pom.xml
```diff
@@ -0,0 +1,417 @@
...
+        </dependency>
+
+        <dependency>
+            <!--Must be JDBC 4 (Java 6) compliant-->
+            <groupId>postgresql</groupId>
+            <artifactId>postgresql</artifactId>
+            <version>${version.postgresql.9}</version>
+        </dependency>
+
+        <!-- dependency>
```
hchiorean: with this commented out, can the tests be run on Oracle ?

hchiorean:
@okulikov with the exception of the questions that are still open, looks good to me

modeshape-persistence-tests/pom.xml
okulikov: @hchiorean , Oracle still a bit tricky, working on this

okulikov:
@hchiorean , @rhauch all tests are working now against all databases. The only minor issue is that Oracle driver should be downloaded and installed manually.

rhauch:
> The only minor issue is that Oracle driver should be downloaded and installed manually.

Why don't we add the JBoss Maven repository that has the JDBC drivers in it to the `settings.xml` file? People outside of JBoss will still have to install the Oracle driver in their Maven repository.

okulikov:
@rhauch , I tried to use Oracle Maven repository with `settings.xml` and `settings-security.xml` but it doesn't work. Let me try other JBoss Maven repository.

okulikov:
@rhauch , Modified `settings.xml` with JBoss Maven repository containing Oracle driver. Now everything smooth.

modeshape-persistence-tests/pom.xml
```diff
@@ -0,0 +1,451 @@
...
+    <name>ModeShape Persistence Tests</name>
+    <description>ModeShape persistence tests</description>
+    <url>http://www.modeshape.org</url>
+    
+    <properties>
+        <database.name>modeshape_db</database.name>
+        <database.user>system</database.user>
+        <database.password>oracle</database.password>
+        
+        <version.mysql.server>5.7</version.mysql.server>
```
rhauch: Shouldn't we define all of the versions in the parent POM?

modeshape-persistence-tests/pom.xml
```diff
@@ -0,0 +1,451 @@
...
+            <!--Must be JDBC 4 (Java 6) compliant-->
+            <groupId>postgresql</groupId>
+            <artifactId>postgresql</artifactId>
+            <version>${version.postgresql.9}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.oracle</groupId>
+            <artifactId>ojdbc6</artifactId>
+            <version>11.2.0.1.0</version>
```
rhauch: Again, shouldn't this be in the parent POM? Perhaps use a property name that signifies its for integration tests, but I still think it needs to be in the parent POM for consistency.

modeshape-persistence-tests/pom.xml
```diff
@@ -0,0 +1,451 @@
...
+                                    <time>120000</time>
+                                </wait>
+                            </run>
+                            <build>
+                                <from>mysql/mysql-server:${version.mysql.server}</from>
+                            </build>
+                        </image>
+                        <image>
+                            <!-- A Docker image using Oracle XE 11g installation mainted by Docker team -->
+                            <!-- name>modeshape/oracle-server-test-database</name -->
```
rhauch: Remove this comment.

modeshape-persistence-tests/pom.xml
rhauch: This does not seem to have been addressed.

modeshape-persistence-tests/pom.xml
```diff
@@ -0,0 +1,451 @@
...
+                                    <goal>integration-test</goal>
+                                </goals>
+                                <configuration>
+                                    <systemPropertyVariables>
+                                        <!-- Make these available to the tests via system properties -->
+                                        <database.driver>org.postgresql.Driver</database.driver>
+                                        <database.url>jdbc:postgresql://${database.hostname}/${database.name}</database.url>
+                                    </systemPropertyVariables>
+                                </configuration>
+                            </execution>
```
rhauch: Don't we also want Oracle in here?

modeshape-persistence-tests/src/test/java/org/modeshape/jcr/ModeshapePersistenceIT.java
```diff
@@ -0,0 +1,2213 @@
...
+            parent.addNode("child2", "jj:content");
+            session.save();
+
+            versionManager.checkpoint(parent.getPath());
+
+            // move to a location under a new parent
+            session.move("/parent/child1", "/parent/_context/child1");
+            session.save();
+
+            // resto
```
rhauch: This test is commented out. Is this intentional?

rhauch:
@okulikov, very thorough and looks great. A few things need to be addressed, though. See my new line comments above, plus there were a few other changes discussed a month ago that aren't yet addressed. Granted, those older discussions don't show up at the bottom of this page, so their harder to see unless you go through the review page from top to bottom.

okulikov:
@rhuach, moved versions to the parent `pom.xml`, removed unused or repeating propreties, added Oracle execution to the `assemby` profile.

rhauch:
@okulikov, still a few todos:
* Rename the `assembly` profile in the new integration test module (nowhere else) per [this comment](https://github.com/ModeShape/modeshape/pull/1580/files#r70430571)
* Remove Mockito if it's not used per [this comment](https://github.com/ModeShape/modeshape/pull/1580/files#r70430360)

@hchiorean What do you think about changing the Travis build to always run the new integration test profile?

okulikov:
@rhauch , @hchiorean , Done. Missed those comments.

modeshape-parent/pom.xml
```diff
@@ -188,7 +188,9 @@
         <version.postgresql.9>9.2-1002.jdbc4</version.postgresql.9>
         <version.postgresql.8>8.4-702.jdbc4</version.postgresql.8>
         <version.mysql>5.1.38</version.mysql>
+        <version.mysql.server>5.7</version.mysql.server>
         <version.com.oracle>11.2.0.4</version.com.oracle>
+        <version.oracle.it.driver>11.2.0.1.0</version.oracle.it.driver>
```
hchiorean: this should be merged with the previous property. Both denote the version of the oracle driver, so we really just want to use 1.

modeshape-parent/pom.xml
```diff
@@ -188,7 +188,9 @@
         <version.postgresql.9>9.2-1002.jdbc4</version.postgresql.9>
         <version.postgresql.8>8.4-702.jdbc4</version.postgresql.8>
         <version.mysql>5.1.38</version.mysql>
+        <version.mysql.server>5.7</version.mysql.server>
```
hchiorean: again, this should be merged with the above property so we only use one version of the driver. Driver versions should be backwards compatible

modeshape-persistence-tests/src/test/resources/cnd/large-collections.cnd
```diff
@@ -0,0 +1,8 @@
+<mix = "http://www.jcp.org/jcr/mix/1.0">
```
hchiorean: is this CND used anywhere ?

modeshape-persistence-tests/src/test/resources/data/cars.xml
```diff
@@ -0,0 +1,46 @@
...
+				<Utility jcr:primaryType="nt:unstructured"  jcr:mixinTypes="mix:referenceable">
+					<Land_x0020_Rover_x0020_LR2 jcr:primaryType="car:Car" car:year="2008" car:msrp="$33,985" car:mpgHighway="23"  car:model="LR2" car:valueRating="5" car:maker="Land_x0020_Rover" car:mpgCity="16" jcr:mixinTypes="mix:referenceable" car:userRating="4"/>
+					<Land_x0020_Rover_x0020_LR3 jcr:primaryType="car:Car" car:year="2008" car:msrp="$48,525" car:mpgHighway="17"  car:model="LR3" car:valueRating="2" car:maker="Land_x0020_Rover" car:mpgCity="12" jcr:mixinTypes="mix:referenceable" car:userRating="5"/>
+					<Hummer_x0020_H3 jcr:primaryType="car:Car" car:year="2008" car:msrp="$30,595" car:mpgHighway="16"  car:model="H3" car:valueRating="4" car:maker="Hummer" car:mpgCity="13" jcr:mixinTypes="mix:referenceable" car:userRating="3"/>
+					<Ford_x0020_F-150 jcr:primaryType="car:Car" car:year="2008" car:msrp="$23,910" car:mpgHighway="20"  car:model="F-150" car:valueRating="1" car:maker="Ford" car:mpgCity="14" jcr:mixinTypes="mix:referenceable" car:userRating="5"/>
+				</Utility>
+			</Cars>
+		</b>
+	</a>
+	<!-- jcr:system jcr:primaryType="mode:system">
```
hchiorean: if this isn't used, it should be removed

hchiorean:
I've added some additional comments for the latest changes.

>@hchiorean What do you think about changing the Travis build to always run the new integration test profile?

@rhauch I think if we want to run this in Travis we should add it as an additional job. I'd like the default job to stay as is (`clean install -Pintegration`) since that means that whenever a new PR is opened one no longer has to locally build and check the PR, but can rely on the Travis build. This is IMO nice to have.

modeshape-parent/pom.xml
okulikov: @hchiorean , I didn't find `11.2.0.4` in any JBoss repository. Can we use `11.2.0.1.0` for both?

hchiorean: absolutely

modeshape-persistence-tests/src/test/resources/data/cars.xml
okulikov: it is used as initial content and referenced in configuration file

okulikov:
@hchiorean , @rhauch ,

The new commit:
- merges driver's version;
- removes version of MySQL docker images from parent `pom.xml` and `latest` tag is now used instead of version tag;
- updates MySQL database URL to prevent warnings about SSL connections;
- removes unused `large-collections.cnd`;
- `cars.xml` still in place since it is used as initial content

modeshape-persistence-tests/pom.xml
```diff
@@ -0,0 +1,442 @@
...
+                                <wait>
+                                    <log>MySQL init process done. Ready for start up.</log>
+                                    <time>120000</time>
+                                </wait>
+                            </run>
+                            <build>
+                                <from>mysql/mysql-server:latest</from>
+                            </build>
+                        </image>
+                        <!-- image>
```
hchiorean: if these images are commented out do we need them here ?

modeshape-persistence-tests/src/test/resources/data/cars.xml
hchiorean: I understand that about the first part of the file; but I'm referring to the commented out around `jcr:versionStorage` which isn't supported anyway

hchiorean:
some more questions:

1. for MySQL why are we using `latest` as the tag ? This seems like a bad idea IMO since we're explicitly depending on a specific version via the driver. What happens when MySQL releases the next version post 5.7 and our driver stops working ? IMO we should explicitly depend on the Docker image version that matches/is compatible with our driver version

2. for Oracle and Postgres it seems that the build doesn't use/configure the docker Maven plugin. There is a configuration section for these 2 databases but it's commented out atm. How does this work ? Similar to (1) IMO we always want Docker to build and use a specific tag, never the latest in order to match our drivers.

thoughts ?

modeshape-persistence-tests/pom.xml
okulikov: That's my bad with commit. I am just commenting images locally to save memory

okulikov:
@hchiorean , I am fine with concrete image versions. Should we define then in parent `pom.xml` versions for the drivers and relative docker images?

hchiorean:
I think the name of the docker images should be defined in the same place where we're defining the driver versions (i.e. the parent pom).

okulikov:
@hchiorean , @rhauch , Updated PR defines versions for docker images and drivers in parent `pom.xml` . 

The latest released version of the driver for MySQL is 5.1.39 which is compatible with 5.7 server version. 

I will also try to redefine profiles with the goal to build and run only dedicated DB image related to the profile (or all for `all` profile)

modeshape-persistence-tests/pom.xml
```diff
@@ -0,0 +1,442 @@
...
+
+    <!-- The groupId and version values are inherited from parent -->
+    <artifactId>modeshape-persistence-tests</artifactId>
+    <packaging>jar</packaging>
+    <name>ModeShape Persistence Tests</name>
+    <description>ModeShape persistence tests</description>
+    <url>http://www.modeshape.org</url>
+    
+    <properties>
+        <database.name>modeshape_db</database.name>
```
hchiorean: these seem to only be used for MySQL and Postgres and not for Oracle. Also, it seems a bit suspect that we're using `system` and `oracle` as values.

hchiorean:
@okulikov: why are we always building and running all docker images as part of the default build instead of doing it part of each profile ? i.e. `mysql` should build and run the `mysql` image; `oracle` should build and run the oracle image; etc

the `all` profile should be the only one which builds and runs all of them. Moreover, stuff like database.username, database.password should be defined as properties per profile IMO, not global for all profiles.

also, as discussed above, the `cars.xml` file contains a section of data which is commented out which doesn't make any sense to me

hchiorean:
note that the plugin configuration regarding the `<image>` section can be placed all together within the modules main `<pluginManagement>` section. It's the actual plugin execution that should be deferred to each profile

okulikov:
@hchiorean , Parent `pom.xml` defines several profiles for DB testing environment including docker. Like this https://github.com/ModeShape/modeshape/blob/master/modeshape-parent/pom.xml#L337 However all of them have a fixed ip address, can I change address on `docker.host` and reuse properties from these profiles for the integration tests as well?

hchiorean:
I think that's a good idea and if we do this, we should go even further: remove all the specific DB profiles from the parent pom and only leave those profiles for which we have a docker image and test in the persistence module.

In other words, for PostgreSql for example: we change the `postgresql8`, `postgresql9` and `postgresql9_local_docker` profiles into one single `postgresql9` profile (which matches the profile from the persistence module)

Similarly, there will be only 1 Oracle and 1 MySql profile and no MsSQL server profile since we haven't really tested anything on this DB. 

The `default_dbms` profile should remain as-is (and default). 

If we make these changes, then we can move the docker manven plugin configuration part (not the execution) to the parent pom's `<pluginManagement>` section

okulikov:
@hchiorean , Actually I already moved docker plugin into parent's pom `pluginManagement` in the follwing way. Docker plagin is defined with 4 executions:
- execution configured for `mysql` image with goals `build` and `start` bound to phase `none`; 
- execution configured for `postgres` image with goals `build` and `start` bound to phase `none`;
- execution configured for `oracle` image with goals `build` and `start` bound to phase `none`;
- execution with goal `stop` bound to phase `post-integration-stop`.

Persistence test module's `pom.xml` defines profiles which inherit docker-maven plugin and override phase of the execution responsible for building and starting image from `none` to `pre-integration-test`. Its working.  

So if this way fine I will continue to merge DB's properties with existing profiles as you mentioned above.

hchiorean:
@okulikov: why do we need to bind defaults for phase `none` ? Wouldn't it be easier to simply not have any executions by default and just the ` <configuration>` section ? Actual executions can then be added only when required in the corresponding modules

okulikov:
@hchiorean , when it is bound to `none` it still not active. Child `pom.xml` actually binds required execution. I am just not sure that it is possible to define `configuration` several times for the same plugin and wrap it with `execution` not bound by default.

hchiorean:
I don't follow: in the parent pom.xml in the `<pluginManagement>` section, there should be the docker plugin configured with just [this section](https://github.com/ModeShape/modeshape/pull/1580/files#diff-59b91671824916c82546ed5ddd9a34edR132).
why would this not work ?

okulikov:
This will build and start all images at the same time. I want to make it configured to build and start only image required by profile, ie `oracle` profile build and starts only Oracle DB.

hchiorean:
ok, maybe I'm missing something: aren't the docker images only built and started during:
```
        <execution>
                  <id>start</id>
                  <phase>pre-integration-test</phase>
                  <goals>
                    <goal>build</goal>
                    <goal>start</goal>
                  </goals>
          </execution>
```
?

I would expect so; I would also expect that a configuration similar to:
```
<pluginManagement>
  <plugin>
              <groupId>io.fabric8</groupId>
              <artifactId>docker-maven-plugin</artifactId>
              <configuration>
                <images>
                  <image></image>
                  <image></image>
              </images>
           </configuration>
      </plugin>
etc
```
does nothing by default.

So I'm proposing that the latter section is part of the parent POM and the former is only defined in the corresponding profiles inside the `persistence-test` module. 

The default DB profiles from the parent `pom.xml` _should not do anything docker-related_ by default. They should only (at most) define some properties which are used elsewhere.

okulikov:
I need to look at the `docker.image` property then.

hchiorean:
the property is mentioned here https://dmp.fabric8.io/#global-configuration

hchiorean:
actually, do we really need to use the `build` part of the plugin at all ? Can't we just configure the plugin to run preexisting docker images (via https://dmp.fabric8.io/#docker:start) and just pass in the username/password environment variables to the image ?

okulikov:
@hchiorean , both features are working nice. I mean `docker.image` property and running preexisting images without definition of the `build` part. Also it is possible (and I think it will be good) to define `executions` for docker plugin in the parent `pom.xml` only once and in persistence test module's `pom.xml` DB specific profile should only set the respective value for the `docker.image` property.

hchiorean:
> Also it is possible (and I think it will be good) to define executions for docker plugin in the parent pom.xml only once and in persistence test module's pom.xml DB specific profile should only set the respective value for the docker.image property.

- if you mean in the `<pluginManagement>` section, then this may be worth a try since it should not impact any modules which don't use the docker plugin
- if you mean in the DB profiles from the parent `pom.xml`, I'm not sure it's a good idea since we don't want Docker running atm. for other modules (maybe we'll change this in the future if necessary)

okulikov:
@hchiorean , pushed new commit with plugins configuration as we discussed above.

hchiorean:
@okulikov: there are couple of things which still remain:
- the `all` profile and all its properties should not be part of the parent pom as no other module is likely to ever use it; as such it should only be in the persistence test module
- the fact that `dataSource.*` properties have been renamed to `database.*` properties, breaks all other files which used this properties (see [this](https://github.com/ModeShape/modeshape/blob/master/persistence/modeshape-persistence-relational/src/test/resources/db-config.json) as an example). As such, all the places where the `dataSource.*` placeholders were being used should be updated as well

modeshape-parent/pom.xml
```diff
@@ -315,76 +318,15 @@
...
-                    <groupId>mysql</groupId>
-                    <artifactId>mysql-connector-java</artifactId>
-                    <version>${version.mysql}</version>
-                    <scope>test</scope>
-                </dependency>
-            </dependencies>
             <properties>
                 <database>mysql5</database>
+                <database.name>mysql</database.name>
                 <dataSource.dialect>org.hibernate.dialect.MySQL5InnoDBDialect</dataSource.dialect>
```
hchiorean: this is not used, and should be removed

modeshape-parent/pom.xml
```diff
@@ -396,80 +338,17 @@
...
-                    <artifactId>postgresql</artifactId>
-                    <version>${version.postgresql.9}</version>
-                    <scope>test</scope>
-                </dependency>
-            </dependencies>
             <properties>
                 <database>postgresq9</database>
-                <dataSource.driverClassName>org.postgresql.Driver</dataSource.driverClassName>
-                <dataSource.url>jdbc:postgresql://db14.mw.lab.eng.bos.redhat.com:5432/dballo04</dataSource.url>
+                <database.name>dballo04</database.name>
```
hchiorean: since we're using docker urls, these 3 properties should be changed from `dballo` to `postgres` really

hchiorean:
@okulikov : see the above 2 comments and also the EDIT to the previous one: I think we should renamed all the dataSource properties to `database`

okulikov:
@hchiorean , we still need to rename properties from `dataSource.` to `database.`? I did this commit before noticed about your comment regardin renaming?

hchiorean:
@okulikov: yes pls, rename the properties (and all affected config files - should be about 4 or 5)

okulikov:
@hchiorean , pushed commit with new names of properties.

hchiorean:
@okulikov: the build is failing (see Travis) so some files were probably not changed.

hchiorean:
btw: `default_dbms` should not be removed as lot of tests depend on it (it's probably the cause of the failure). It should remain as-is.

okulikov:
@hchiorean , I did return `default_dbms` profile. It was definitely the reason for failure.

modeshape-parent/pom.xml
```diff
@@ -187,10 +187,14 @@
         <version.com.h2>1.4.191</version.com.h2>
         <version.postgresql.9>9.2-1002.jdbc4</version.postgresql.9>
         <version.postgresql.8>8.4-702.jdbc4</version.postgresql.8>
-        <version.mysql>5.1.38</version.mysql>
-        <version.com.oracle>11.2.0.4</version.com.oracle>
+        <version.mysql>5.1.39</version.mysql>
+        <version.com.oracle>11.2.0.1.0</version.com.oracle>
         <version.com.microsoft.sqlserver>4.0.2206</version.com.microsoft.sqlserver>
```
hchiorean: is suspect this is not used anymore, so it should be removed

hchiorean:
@okulikov: just one minor comment (see above), the rest of this looks fine to me; nice work
@rhauch: let me know if/when u'r ok with merging this

rhauch:
@okulikov, nice work on this new module. @hchiorean, once the SQL Server Maven property is cleaned up, this can be merged.

hchiorean:
@okulikov  after merging https://github.com/ModeShape/modeshape/pull/1586, this can no longer be merged cleanly; pls update this PR after pulling the latest `master` and rebasing; thanks

okulikov:
@hchiorean , Done.

hchiorean:
@okulikov: tried running the persistence module locally but it seems the Oracle Maven driver cannot be located:
```
[ERROR] Failed to execute goal on project modeshape-persistence-tests: Could not resolve dependencies for project org.modeshape:modeshape-persistence-tests:jar:5.2-SNAPSHOT: Could not find artifact co
m.oracle:ojdbc6:jar:11.2.0.1.0 in earlyaccess (http://maven.repository.redhat.com/earlyaccess/all) -> [Help 1]
[ERROR]
```

searching [JBoss Nexus](https://repository.jboss.org/nexus/#nexus-search;quick~ojdbc6) yields no results. Where is this artifact supposed to be located ? Do we need an external Maven repo ?

Stumbling across this issue while doing a simple `clean install` on the persistence-tests module, I noticed another issue: the driver artifact dependencies are listed only in this module's global `<dependencies>` section. Instead, the driver dependencies should _be moved to the parent pom_ in each of the specific DB profiles section.

So overall there are still 2 issues:
* the location of the Oracle driver in Maven
* the fact that the driver artifact dependencies should be listed as dependencies in each of the parent DB profiles

hchiorean:
@okulikov  I suspect that to get Oracle running one needs to install manually the driver locally; if that's the case, then the `all` profile should be modified and Oracle excluded since it requires a manual step.

Oracle should still be able to run by explicitly running the `oracle` profile

okulikov:
@hchiorean , the artifact located in this [repository] (https://repository.jboss.org/nexus/content/repositories/public-jboss/)
The respective [modification] (https://github.com/ModeShape/modeshape/pull/1580/files#diff-4d5199f7e02379b759916edeaf048f64R64) included into the `settings.xml`


I will update dependencies regarding profiles

hchiorean:
@okulikov: thanks, I missed that change

hchiorean:
@okulikov: running `clean install` on the `modeshape-persistence-tests` modules (which postgres the default profile) fails with:
```
[ERROR] Failed to execute goal io.fabric8:docker-maven-plugin:0.14.1:start (docker-postgres) on project modeshape-persistence-tests: Execution docker-postgres of goal io.fabric8:docker-maven-plugin:0.
14.1:start failed: Failed to create contained configuration for [postgres:9.5.4]: Env variable 'POSTGRES_DB' must not be null or empty -> [Help 1]
[ERROR]
```

hchiorean:
@okulikov: the problem comes from the fact that there are 2 different activeByDefault profiles:
```
 - postrgresql9 (source: org.modeshape:modeshape-persistence-tests:5.2-SNAPSHOT)
 - default_dbms (source: org.modeshape:modeshape-parent:5.2-SNAPSHOT)
```

So we have 2 options IMO:
1. add a `default_dbms` profile to the persistence which uses H2 and no docker. Depending on how the tests cleanup stuff this may or may not work OOTB 
2. add a `default_dbms` profile to the persistence module which (re)defines Postgresl9 properties. In other words for this module the `default_dbms` is not H2, but Postgresql

okulikov:
@hchiorean , last commit redefines default `default_dms` profile with `Postgres` properties. On my side it is working. Dependencies also moved to parent `pom.xml` into specific profile section.

hchiorean:
@okulikov : this looks good; one final thing though: can you pls add a profile to the root (not parent) pom which includes this modules (e.g. `persistence-integration`)

okulikov:
@hchiorean , Done.

PR #1581, opened 2016-07-15
MODE-2077 Adds the ability to use DB locking when running in a cluster instead of JGroups locking

hchiorean:
This can be configured in JSON like so:
```
"clustering" : {
        "clusterName" : "Test cluster",
        "configuration" : "config/cluster/jgroups-test-config.xml",
        "locking" : "db"
    },
```
or in Wildfly:

```
<repository name="repo-clustered3 cluster-name="modeshape-wf-it2" cluster-stack="tcp" cluster-locking="db">
```

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/JcrRepository.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/JcrRepository.java
@@ -100,7 +100,8 @@ import org.modeshape.jcr.federation.FederatedDocumentStore;
 import org.modeshape.jcr.journal.ChangeJournal;
 import org.modeshape.jcr.journal.ClusteredJournal;
 import org.modeshape.jcr.journal.LocalJournal;
-import org.modeshape.jcr.locking.ClusteredLockingService;
+import org.modeshape.jcr.locking.DbLockingService;
+import org.modeshape.jcr.locking.JGroupsLockingService;
 import org.modeshape.jcr.locking.LockingService;
 import org.modeshape.jcr.locking.StandaloneLockingService;
 import org.modeshape.jcr.mimetype.MimeTypeDetector;
@@ -1051,12 +1052,15 @@ public class JcrRepository implements org.modeshape.jcr.api.Repository {
                         } else {
                             this.clusteringService = ClusteringService.startStandalone(clusterName, clustering.getConfiguration());        
                         }
+                        JGroupsLockingService jgroupsLockingService = new JGroupsLockingService(this.clusteringService.getChannel(), 
+                                                                                                config.getLockTimeoutMillis());
+                        this.lockingService = clustering.useDbLocking() ? 
+                                              new DbLockingService(jgroupsLockingService, RepositoryCache.INITIALIZATION_LOCK,
+                                                                   this.schematicDb) : jgroupsLockingService;
                     } else {
                         this.clusteringService = null;
+                        this.lockingService =  new StandaloneLockingService(config.getLockTimeoutMillis());
                     }
-                    this.lockingService = this.clusteringService != null ? 
-                                          new ClusteredLockingService(this.clusteringService.getChannel(), config.getLockTimeoutMillis()) : 
-                                          new StandaloneLockingService(config.getLockTimeoutMillis());
                   
                     suspendExistingUserTransaction();
                     
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/JcrRepository.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/JcrRepository.java
@@ -1044,6 +1044,8 @@ public class JcrRepository implements org.modeshape.jcr.api.Repository {
                     this.connectors = new Connectors(this, connectorComponents, extSources, preconfiguredProjectionsByWorkspace);                    
                    
                     RepositoryConfiguration.Clustering clustering = config.getClustering();
+                    LockingService startupLockingService = null;
+                    long lockTimeoutMillis = config.getLockTimeoutMillis();
                     if (clustering.isEnabled()) {
                         final String clusterName = clustering.getClusterName();
                         Channel channel = environment().getChannel(clusterName);
@@ -1052,14 +1054,12 @@ public class JcrRepository implements org.modeshape.jcr.api.Repository {
                         } else {
                             this.clusteringService = ClusteringService.startStandalone(clusterName, clustering.getConfiguration());        
                         }
-                        JGroupsLockingService jgroupsLockingService = new JGroupsLockingService(this.clusteringService.getChannel(), 
-                                                                                                config.getLockTimeoutMillis());
+                        startupLockingService = new JGroupsLockingService(this.clusteringService.getChannel(), lockTimeoutMillis);
                         this.lockingService = clustering.useDbLocking() ? 
-                                              new DbLockingService(jgroupsLockingService, RepositoryCache.INITIALIZATION_LOCK,
-                                                                   this.schematicDb) : jgroupsLockingService;
+                                              new DbLockingService(lockTimeoutMillis, this.schematicDb) : startupLockingService;
                     } else {
                         this.clusteringService = null;
-                        this.lockingService =  new StandaloneLockingService(config.getLockTimeoutMillis());
+                        this.lockingService =  new StandaloneLockingService(lockTimeoutMillis);
                     }
                   
                     suspendExistingUserTransaction();
@@ -1113,8 +1113,7 @@ public class JcrRepository implements org.modeshape.jcr.api.Repository {
                     this.documentStore = connectors.hasConnectors() ? new FederatedDocumentStore(connectors, localStore) : localStore;
 
                     // Set up the repository cache ...
-                    LockingService cacheLockingService = clusteringService != null ? lockingService : null;
-                    this.cache = new RepositoryCache(context, documentStore, cacheLockingService, config, systemContentInitializer,
+                    this.cache = new RepositoryCache(context, documentStore, startupLockingService, config, systemContentInitializer,
                                                      repositoryEnvironment, changeBus, Upgrades.STANDARD_UPGRADES);
 
                     // Set up the node type manager ...
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/locking/DbLockingService.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/locking/DbLockingService.java
@@ -1,0 +1,86 @@
+/*
+ * ModeShape (http://www.modeshape.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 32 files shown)


modeshape-jcr/src/main/java/org/modeshape/jcr/JcrRepository.java
```diff
@@ -1051,12 +1052,15 @@ protected RunningState( JcrRepository.RunningState other,
                         } else {
                             this.clusteringService = ClusteringService.startStandalone(clusterName, clustering.getConfiguration());        
                         }
+                        JGroupsLockingService jgroupsLockingService = new JGroupsLockingService(this.clusteringService.getChannel(), 
+                                                                                                config.getLockTimeoutMillis());
+                        this.lockingService = clustering.useDbLocking() ? 
+                                              new DbLockingService(jgroupsLockingService, RepositoryCache.INITIALIZATION_LOCK,
+                                                                   this.schematicDb) : jgroupsLockingService;
```
rhauch: Curious why `JGroupsLockingService` instance is always created before knowing whether it's used, while the `DbLockingService` is not. My only naive concern would be whether the `JGroupsLockingService` obtains resources on creation, and if so then it may be doing this unnecessarily.

rhauch: Ah, never mind. I now see that the `DbLockingService` uses the `JGroupsLockingService` instance.

modeshape-jcr/src/main/java/org/modeshape/jcr/locking/DbLockingService.java
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Creates a new db locking service instance.
+     * 
+     * @param jgroupsLockingService the {@link JGroupsLockingService} instance to which some locking may be delegated; never {@code null}
+     * @param initializationLockName the name of the lock used for cluster-wide initialization, which cannot be handled by this 
+     * service; never {@code null}
+     * @param db a {@link Lockable} instance; never {@code null}
+     */
+    public DbLockingService(JGroupsLockingService jgroupsLockingService, String initializationLockName, Lockable db) {
```
rhauch: Does this need to be a `JGroupsLockingService`, or could it just be any `LockingService` instance?

modeshape-jcr/src/main/java/org/modeshape/jcr/locking/DbLockingService.java
```diff
@@ -0,0 +1,86 @@
...
+        this.jgroupsLockingService = jgroupsLockingService;
+        this.initializationLockName = initializationLockName;
+        this.db = db;
+    }
+
+    @Override
+    public boolean tryLock(long time, TimeUnit unit, String... names) throws InterruptedException {
+        if (isInitializationLock(names)) {
+            // the db cannot handle the initialization lock, so delegate to JG
+            return jgroupsLockingService.tryLock(time, unit, names[0]);
```
rhauch: It seemed really weird that the DB locking service would rely upon the JGroups locking service (and the JavaDoc didn't help), until I looked at this line. I'm *assuming* that the repository needs some locking service before the database locking service can be initialized -- is that right? I'd suggest expanding upon this in the JavaDoc for the class.

modeshape-jcr/src/main/java/org/modeshape/jcr/locking/DbLockingService.java
```diff
@@ -0,0 +1,86 @@
...
+        return result;        
+    }
+
+    private boolean isInitializationLock(String... names) {
+        return names.length == 1 && names[0].equals(initializationLockName);
+    }
+
+    @Override
+    public boolean tryLock(String... names) throws InterruptedException {
+        return tryLock(jgroupsLockingService.getLockTimeoutMillis(), TimeUnit.MILLISECONDS, names);
```
rhauch: Why use the lock timeout length from the JGroups locking service?

persistence/modeshape-persistence-relational/src/main/java/org/modeshape/persistence/relational/DefaultStatements.java
```diff
@@ -297,21 +332,20 @@ public void remove( List<String> ids ) throws SQLException {
             runBatchOperation(connection, sql, ids, batchLoadSize(), this::batchRemove);    
         }
 
-        private void batchRemove( Connection connection, String statement, int startIdx, int endIdx, List<String> ids )
-                throws SQLException {
-            List<String> sublist = ids.subList(startIdx, endIdx);
-            String params = sublist.stream().map(id -> "?").collect(Collectors.joining(","));
+        private boolean batchRemove( Connection connection, String statement, List<String> ids ) throws SQLException {
+            String params = ids.stream().map(id -> "?").collect(Collectors.joining(","));
             String statementString = statement.replaceAll("#", params);
```
rhauch: Lines 336 and 337 are used in several places (at least 3). Would it be useful to extract to a helper method? It'd probably help with understanding the code; took me a moment to figure out what they're doing.

persistence/modeshape-persistence-relational/src/main/resources/org/modeshape/persistence/relational/sqlserver_database.properties
```diff
@@ -30,4 +30,7 @@ update_content = UPDATE {0} SET LAST_CHANGED=CURRENT_TIMESTAMP, CONTENT = ? WHER
 remove_content = DELETE FROM {0} WHERE ID IN (#)
 
 # Remove all documents
-remove_all_content = DELETE FROM {0}
\ No newline at end of file
+remove_all_content = DELETE FROM {0}
+
+# Lock documents
+lock_content = SELECT ID FROM {0} WHERE ID IN (#) WITH (UPDLOCK)
```
rhauch: Should `ROWLOCK` hint also be used here in addition to `UPDLOCK`? From the [documentation on SQL Server hints](https://msdn.microsoft.com/en-us/library/ms187373.aspx):

> UPDLOCK
> Specifies that update locks are to be taken and held until the transaction completes. UPDLOCK takes update locks for read operations only at the row-level or page-level....

and

> ROWLOCK
> Specifies that row locks are taken when page or table locks are ordinarily taken. When specified in transactions operating at the SNAPSHOT isolation level, row locks are not taken unless ROWLOCK is combined with other table hints that require locks, such as UPDLOCK and HOLDBACK.

So if the database is using SNAPSHOT isolation level, then an `UPDLOCK` hint doesn't do anything.

> SNAPSHOT
> ... The transaction can only recognize data modifications that were committed before the start of the transaction. Data modifications made by other transactions after the start of the current transaction are not visible to statements executing in the current transaction. The effect is as if the statements in a transaction get a snapshot of the committed data as it existed at the start of the transaction.

So if SNAPSHOT isolation were used with MySQL, then using just `UPDLOCK` means that no locking actually happens, and coupled with SNAPSHOT isolation means that two competing transactions might both read a node's state, the second txn to commit an update will overwrite the first txn's update and the first transaction's changes *will be lost*. However, using both the `UPDLOCK,ROWLOCK` hints would cause the database to acquire locks even when using SNAPSHOT isolation, meaning the transactions get serialized.

modeshape-jcr/src/main/java/org/modeshape/jcr/locking/DbLockingService.java
hchiorean: it could be any instance, for abstraction reasons. For practical reasons since we're enforcing the fact that this should only be used in a cluster it has to be JGroups as it's the only other cluster locking service.

modeshape-jcr/src/main/java/org/modeshape/jcr/locking/DbLockingService.java
hchiorean: ModeShape uses an ephemeral lock for initialization which does not correspond to a DB entry. Therefore we can't use the DB locking service for this lock.

rhauch: Please add this to the JavaDoc for the class, and even the JavaDoc for the constructor parameter.

modeshape-jcr/src/main/java/org/modeshape/jcr/locking/DbLockingService.java
hchiorean: Because it's the timeout coming from the JSON configuration (locking timeout can be configured). Of course the other option is to duplicate the 'timeout' member in this instance and make sure `RepositoryConfiguration` sets it.

persistence/modeshape-persistence-relational/src/main/java/org/modeshape/persistence/relational/DefaultStatements.java
hchiorean: makes sense, sure

persistence/modeshape-persistence-relational/src/main/resources/org/modeshape/persistence/relational/sqlserver_database.properties
hchiorean: ok, I can change this to UPLOCK and ROWLOCK. 

Note that the SQLServer support is really generic, since there's no way for me to actually test it :)

hchiorean:
amended the previous commit not only with some of the discussed changes, but also with a slightly different code for the `DBLockingService` (after some thought, there is a much cleaner way of doing things than to couple the DB and JGroups locking services together)

modeshape-jcr/src/main/java/org/modeshape/jcr/locking/DbLockingService.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    public boolean tryLock(long time, TimeUnit unit, String... names) throws InterruptedException {
+        long start = System.nanoTime();
+        boolean result = false;
+        while (!(result = db.lockForWriting(names)) && 
+               TimeUnit.MILLISECONDS.convert((System.nanoTime() - start), TimeUnit.NANOSECONDS) < time) {
+            //wait a bit
+            Thread.sleep(1000);
```
rhauch: Do we really want to sleep for 1 sec? Also, with a constant sleep time it is possible that one thread gets starved of locks, if other threads keep jumping in. Might be worth sleeping for a random amount of time, say between 50 and 300 ms.

Also, you're not converting `time` to ns here.

persistence/modeshape-persistence-relational/src/main/java/org/modeshape/persistence/relational/RelationalDb.java
```diff
@@ -197,6 +197,15 @@ public Document get(String key) {
     }
 
     @Override
+    public boolean lockForWriting( List<String> locks ) {
+        TransactionsHolder.requireActiveTransaction();
+        if (locks.isEmpty()) {
+            return false;
+        }
```
rhauch: Should this `if` block be before the `requireActiveTransaction()` call?

rhauch:
A few comments, but otherwise looks good. I like the way the repository is using a separate startup locking service, and how that simplifies the locking service used once the cluster is started.

modeshape-jcr/src/main/java/org/modeshape/jcr/locking/DbLockingService.java
hchiorean: I agree with the randomization of the sleep duration, but I'm not sure I understand the other comment: the `time` param is in ms and I'm converting everything to `ms`, even though I'm using `ns` for duration. So the comparison should be `ms` against `ms`

hchiorean:
amended the PR

rhauch:
Looks great.

PR #1583, opened 2016-07-25
MODE-2617 Fixes the cache retrieval of new versions

hchiorean:
Under certain timing conditions (visible especially when running in a cluster) a newly created version could be reported as not found since the events used to clear the cache are async

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/JcrVersionManager.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/JcrVersionManager.java
@@ -374,8 +374,9 @@ final class JcrVersionManager implements org.modeshape.jcr.api.version.VersionMa
         } finally {
             // TODO: Versioning: may want to catch this block and retry, if the new version name couldn't be created
         }
-
-        return (JcrVersionNode)session.node(version, Type.VERSION);
+        // return the version node from the system cache, not this session's cache because this session is asynchronously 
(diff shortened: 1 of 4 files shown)


modeshape-jcr/src/main/java/org/modeshape/jcr/JcrVersionManager.java
```diff
@@ -374,8 +374,9 @@ JcrVersionNode checkin( AbstractJcrNode node ) throws RepositoryException {
         } finally {
             // TODO: Versioning: may want to catch this block and retry, if the new version name couldn't be created
         }
-
-        return (JcrVersionNode)session.node(version, Type.VERSION);
+        // return the version node from the system cache, not this session's cache because this session is asynchronously 
+        // notified of system ws changes and may not always get the latest data immediately
+        return (JcrVersionNode)session.node(version, systemSession, Type.VERSION, null);
```
rhauch: I know this relates to versioning, but might we miss out on changes that we've made and haven't yet committed? Or do we always want the version manager to read the data from the system session.

hchiorean:
version nodes always reside in the system area; this particular method returns a JCR view of the version node, so it should do so by using the "system workspace cache", not the cache from the regular JCR session

the rest of the versioning API as far as I can tell doesn't exhibit this problem so in this particular case it was an issue with the `checkin` method

rhauch:
@hchiorean, looks good.

PR #1584, opened 2016-08-17
Resolves https://issues.jboss.org/browse/MODE-2622:

dbernstein:
unit tests in modeshape-jcr breaking on OSX + java 1.8.0_20


hchiorean:
@dbernstein: I think this could be improved a bit to avoid iterating `localHostAddresses` twice: during the first iteration hold onto the first non `Inet6Address` via a separate variable and then after the loop if `localHost` is null, simply return that. wdyt ?

dbernstein:
Sure I could do that.

PR #1589, opened 2016-09-01
MODE-2626 Fixes the custom binary storage integration with the WF kit

hchiorean:


deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/AddCustomBinaryStorage.java
```diff
@@ -39,17 +45,28 @@ protected void writeBinaryStorageConfiguration( String repositoryName,
...
+    
+    private void writeCustomProperty(EditableDocument binaries, Property property) {
+        ModelNode value = property.getValue();
+        if (!value.isDefined()) {
+            return;
         }
+        String name = property.getName();
+        if (COMMON_BINARY_ATTRIBUTES.contains(name)) {
+            // these are set via the base class...
+            return;
```
rhauch: How does this work to set them via the base class? This method doesn't call 'super'.

hchiorean: via [this method|https://github.com/ModeShape/modeshape/blob/master/deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/AbstractAddBinaryStorage.java#L101] called first when initializing any binary store.

rhauch: Thanks!

rhauch:
@hchiorean, looks great.

PR #1590, opened 2016-09-02
MODE-2623 Fixes the loading of documents from the persistent store after exclusive locks are obtained

hchiorean:
The previous code always loaded a fresh copy of the latest persisted (i.e. successfully committed) information and did not take into account the fact that a user transaction might span multiple workspace cache usages. 

This is particularly evident when dealing with the `jcr:system` area and `SystemContent` in general, via multiple `session.save()` calls coming from different API methods.

### DIFF
--- a/persistence/modeshape-persistence-file/src/main/java/org/modeshape/persistence/file/FileDb.java
+++ b/persistence/modeshape-persistence-file/src/main/java/org/modeshape/persistence/file/FileDb.java
@@ -102,13 +102,13 @@ public class FileDb implements SchematicDb {
 
     @Override
     public List<SchematicEntry> load( Collection<String> keys ) {
+        final TransactionStore.TransactionMap<String, Document> txContent = transactionalContent(false);
+        final TransactionStore.TransactionMap<String, Document> actualContent = txContent != null ? txContent : persistedContent; 
         return keys.stream()
-                   .map(persistedContent::get)
+                   .map(actualContent::get)
                    .filter(Objects::nonNull)
-                   .map(document -> {
-                       SchematicEntry schematicEntry = () -> document;
-                       return schematicEntry;
-                   }).collect(Collectors.toList()); 
+                   .map(SchematicEntry::fromDocument)
+                   .collect(Collectors.toList()); 
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


persistence/modeshape-persistence-file/src/main/java/org/modeshape/persistence/file/FileDb.java
```diff
@@ -102,13 +102,13 @@ public Document get( String key ) {
...
+        final TransactionStore.TransactionMap<String, Document> actualContent = txContent != null ? txContent : persistedContent; 
         return keys.stream()
-                   .map(persistedContent::get)
+                   .map(actualContent::get)
                    .filter(Objects::nonNull)
-                   .map(document -> {
-                       SchematicEntry schematicEntry = () -> document;
-                       return schematicEntry;
-                   }).collect(Collectors.toList()); 
+                   .map(document -> (SchematicEntry) () -> document)
```
rhauch: It is possible to use `.map(SchematicEntry.class::class)` instead. Not sure if that's more readable, though.

rhauch: Ok, scratch that. Took me a minute that you're not casting but instead are creating a `SchematicEntry` using this technique. Since that operation is done in several places, it's probably useful to add a static method to `SchematicEntry` that is something like:

```
public static SchematicEntry fromDocument(Document source) {
    return ()->source;
}
```

Then the above line could be written:

```
.map(SchematicEntry::fromDocument)
```

which I think is more readable.

persistence/modeshape-persistence-relational/src/main/java/org/modeshape/persistence/relational/RelationalDb.java
```diff
@@ -173,26 +174,37 @@ public Document get(String key) {
...
+                if (alreadyWrittenForTx != null) {
+                    // remove the key so we don't load it again from the DB
+                    keysIterator.remove();
+                    alreadyChangedInTransaction.add(() -> alreadyWrittenForTx);
+                }
+            }
+        }
+
         Function<Document, SchematicEntry> documentParser = document -> {
             SchematicEntry entry = () -> document;
```
rhauch: Again, I think the following would make this easier to read:

```
SchematicEntry entry = SchematicEntry.fromDocument(document);
```

rhauch:
Looks great, though I have a suggestion above.

hchiorean:
amended the commit

rhauch:
Thanks. Looks good to me.

PR #1591, opened 2016-09-06
MODE-2627 Fixes additional transactional issues around the workspace cache

hchiorean:
Each workspace cache should be cleared individually when a transaction commits. The previous code only cleared the first ws cache which registered a transaction synchronization


rhauch:
Looks good.

PR #1592, opened 2016-09-07
MODE-2628 Fixes the MongoDB configuration in the WF kit

hchiorean:
In addition, adds the Java drivers for Mongo, Cassandra and S3 to the BOMs


rhauch:
Looks good.

PR #1593, opened 2016-09-16
MODE-2629: mitigate Oracle 11g query syntax errors.

dnillia:
- mitigate the problem where Oracle 11g refuses to execute a query that has more than
  1,000 parameters passed into the IN clause; the problem is mitigated, but not fixed;
  it may still be possible to run into problems if Oracle 11g has limitations as to how
  many OR clauses a query can have


hchiorean:
superseded by https://github.com/ModeShape/modeshape/pull/1597; thanks for helping out with this issue @dnillia

PR #1594, opened 2016-09-20
MODE-2566 Upgrades to OpenCMIS 1.0.0 and explicitly advertises CMIS 1.1 as the supported version for repositories

hchiorean:


rhauch:
@hchiorean, looks great.

PR #1595, opened 2016-09-21
MODE-2620: Method node.getParent() throws NullPointerException

okulikov:
This PR fixes silent exception handling of the failed transactions what causes repository inconsistency.

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/WritableSessionCache.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/WritableSessionCache.java
@@ -570,15 +570,14 @@ public class WritableSessionCache extends AbstractSessionCache {
                 } catch (IllegalStateException err) {
                     // Not associated with a txn??
                     throw new SystemFailureException(err);
-                } catch (RollbackException err) {
+                } catch (RollbackException | HeuristicMixedException | HeuristicRollbackException err) {
                     // Couldn't be committed, but the txn is already rolled back ...
-                    return;
-                } catch (HeuristicMixedException err) {
-                    // Rollback has occurred ...
-                    return;
-                } catch (HeuristicRollbackException err) {
-                    // Rollback has occurred ...
-                    return;
+                    txn = null;
+                    if (err.getCause() != null) {
+                        throw err.getCause();
+                    } else {
+                        throw new SystemFailureException(err);
+                    }
                 } catch (SystemException err) {
                     // System failed unexpectedly ...
                     throw new SystemFailureException(err);
@@ -767,13 +766,15 @@ public class WritableSessionCache extends AbstractSessionCache {
                 } catch (IllegalStateException err) {
                     // Not associated with a txn??
                     throw new SystemFailureException(err);
-                } catch (RollbackException err) {
+                } catch (RollbackException | HeuristicMixedException | HeuristicRollbackException err) {
                     // Couldn't be committed, but the txn is already rolled back ...
-                    return;
-                } catch (HeuristicMixedException err) {
-                } catch (HeuristicRollbackException err) {
-                    // Rollback has occurred ...
-                    return;
+                    txn = null;
+                    if (err.getCause() != null) {
+                        throw err.getCause();
+                    } else {
+                        throw new SystemFailureException(err);
+                    }
+                    //throw new SystemFailureException(err);
                 } catch (SystemException err) {
                     // System failed unexpectedly ...
                     throw new SystemFailureException(err);
@@ -785,8 +786,8 @@ public class WritableSessionCache extends AbstractSessionCache {
 
         } catch (RuntimeException e) {
             throw e;
-        } catch (Exception e) {
-            throw new WrappedException(e);
+        } catch (Throwable t) {
+            throw new WrappedException(t);
         } finally {
             try {
                 thatLock.unlock();
@@ -913,13 +914,14 @@ public class WritableSessionCache extends AbstractSessionCache {
                 } catch (IllegalStateException err) {
                     // Not associated with a txn??
                     throw new SystemFailureException(err);
-                } catch (RollbackException err) {
+                } catch (RollbackException | HeuristicMixedException | HeuristicRollbackException err) {
                     // Couldn't be committed, but the txn is already rolled back ...
-                    return;
-                } catch (HeuristicMixedException err) {
-                } catch (HeuristicRollbackException err) {
-                    // Rollback has occurred ...
-                    return;
+                    txn = null;
+                    if (err.getCause() != null) {
+                        throw err.getCause();
+                    } else {
+                        throw new SystemFailureException(err);
+                    }
                 } catch (SystemException err) {
                     // System failed unexpectedly ...
                     throw new SystemFailureException(err);
@@ -931,8 +933,8 @@ public class WritableSessionCache extends AbstractSessionCache {
 
         } catch (RuntimeException e) {
             throw e;
-        } catch (Exception e) {
-            throw new WrappedException(e);
+        } catch (Throwable t) {
+            throw new WrappedException(t);
         } finally {
             try {
                 thatLock.unlock();
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/Connectors.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/Connectors.java
@@ -294,11 +294,10 @@ public final class Connectors {
         }
     }
 
-    private void storeProjection( Projection projection ) {
+    private void storeProjection( Projection projection, SessionCache systemSession ) {
         PropertyFactory propertyFactory = repository.context().getPropertyFactory();
 
         // we need to store the projection mappings so that we don't loose that information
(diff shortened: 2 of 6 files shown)


hchiorean:
first, I'm not sure I understand how `RollbackException` during `session.save()` is causing federation to fail...

second, `RollbackException` indicates that the tx has been marked for rollback and if there are no other bugs, all the synchronizations should have cleaned up/reverted their state. As such, why should the exception be rethrown ? A transaction is marked for rollback if another issue (i.e. exception) is raised while that transaction is active by some piece of code. 

I'm only referring to `RollbackException` because all the `Heuristic**` ones can never happen in ModeShape since distributed transactions are not supported.

okulikov:
During creation of projection `save()` is called [here](https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/main/java/org/modeshape/jcr/Connectors.java#L312) and [here](https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/main/java/org/modeshape/jcr/Connectors.java#L331). If first call was without problems and second fails then projection will exist but relation between projected and external key will be not persisted but not broken yet. This relation will be broken after restart and then when calling `node.getParent()` [this call] (https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/main/java/org/modeshape/jcr/JcrNode.java#L67) will return `null` as a value for the `parentKey` varibale and later throw NPE 


`java.lang.NullPointerException
	at com.googlecode.concurrentlinkedhashmap.ConcurrentHashMapV8.get(ConcurrentHashMapV8.java:934)
	at com.googlecode.concurrentlinkedhashmap.ConcurrentLinkedHashMap.get(ConcurrentLinkedHashMap.java:672)
	at org.modeshape.jcr.cache.document.WorkspaceCache.getNode(WorkspaceCache.java:233)
	at org.modeshape.jcr.cache.document.AbstractSessionCache.getNode(AbstractSessionCache.java:177)
	at org.modeshape.jcr.cache.document.WritableSessionCache.getNode(WritableSessionCache.java:200)
	at org.modeshape.jcr.JcrSession.node(JcrSession.java:491)
	at org.modeshape.jcr.JcrSession.node(JcrSession.java:476)
	at org.modeshape.jcr.JcrNode.getParent(JcrNode.java:69)
	at org.modeshape.jcr.JcrNode.getParent(JcrNode.java:36)`

hchiorean:
it seems to me that there shouldn't be 2 `session.save` calls but instead just 1 since that entire method should be atomic if there isn't a `mode:federation` node. Did you try simply removing the 1st `session.save` call ? Does that fix the problem ?

>If first call was without problems and second fails then projection will exist but relation between projected and external key will be not persisted but not broken yet. 

if that is the case, then according to the code the `mode:federation` node under `jcr:system` should exist but should not have any children, right ? So there shouldn't be any relation persisted since the exception will bubble up to the caller

>This relation will be broken after restart and then when calling node.getParent() this call will return null as a value for the parentKey varibale and later throw NPE 

which node has a `null` parentKey ? If the 1st `session.save` call was successful then `mode:federation` should have `jcr:system` as parent. If the 2nd `session.save` failed, then this exception should've bubbled up and prevented anything from being stored.

If the exception for the 2nd `session.save` doesn't bubble up, why is that ?

okulikov:
I agree that there should be only one `session.save()` call. I didn't test this way but I think it will fix the issue as well.
> if that is the case, then according to the code the mode:federation node under jcr:system should 
exist but should not have any children, right ?



Yes. 

> which node has a null parentKey ?



External node. Because information about its parent was not persisted as it should (as children node under `mode:federation`)

> If the exception for the 2nd session.save doesn't bubble up, why is that ?


Currently because it is handled [silently](https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/WritableSessionCache.java#L576) and this PR propose to fire exception always.

hchiorean:
>Currently because it is handled silently and this PR propose to fire exception always.

the proposal is to handle a `RollbackException`. But for that to be raised by the transaction manager, it means that another exception has to occur (at least that's my understanding on the JTA API). So if my understanding is correct, another exception is raised somewhere, which causes the tx manager to throw a `RollbackException`. What is this exception and why is it happening ?

okulikov:
>But for that to be raised by the transaction manager, it means that another exception has to occur


I guess so as well but unfortunately still not able to catch the root exception. This might be even some kind of timeout since transaction is not rolled back regular but occasionally. So the proposed PR is just a minimal thing.

hchiorean:
what about changing the code so that there is only one `session.save` at the end as opposed to the 2 calls ? Does the problem still show up ?

okulikov:
I am completely agree about merging two `session.save` calls but it is not enough to make creation of projection atomic.  [Here] (https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/main/java/org/modeshape/jcr/ModeShapeFederationManager.java#L76) is another `session.save()` which is a part of projection procedure. So exception bubble still mandatory or we need to try definitely make operation atomic.

hchiorean:
ok, so it seems to me the entire `createProjection` method has to be atomic, even though it's essentially using 2 session caches - a regular one in this method and a system one in the `Connectors#storeProjection` method.

So how about if we change the code so that:
* both session caches are created in the top level `createProjection` method
* the system session cache is passed _all the way down_ as a parameter to the `Connectors#storeProjection` method 
* the `Connectors#storeProjection` does not call `save` at all anymore
* the `createProjection` method calls `sessionCache.save(systemCache, null)` at the end

This would seem to ensure atomicity for the entire operation

okulikov:
looks fine to me, let me do it

hchiorean:
@okulikov: looks good to me, with one minor comment (see above)

okulikov:
@hchiorean , fixed order. Do you think to leave two commits separated (for me they looks like logically separated)

hchiorean:
@okulikov: I didn't see the 1st commit is still present: is that still required ? I thought that once the `createProjection` operation is atomic there should no need to bubble `RollbackException` up

okulikov:
@hchiorean , I think exception should bubble in all cases at least just for notification. But let me try to simulate what will happen in case of tx failure in details.

hchiorean:
@okulikov any update on this ?

okulikov:
@hchiorean , just a couple of hours for finish. Testing with random exceptions

okulikov:
@hchiorean , exception handling is still required. For instance, If we don't bubble up `RollbackException`  then [creating projection](https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/test/java/org/modeshape/jcr/JcrRepositoryStartupTest.java#L167) completes successful but requesting projected [node] (https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/test/java/org/modeshape/jcr/JcrRepositoryStartupTest.java#L169) will fail with `PathNotFoundException`. This because transaction was rolled back but there is no any kind of indications at the top level.

hchiorean:
@okulikov : how about bubbling up the cause of `RollbackException` ? (i.e. `getCause`) IMO that should be the original exception

okulikov:
@hchiorean , sounds good

okulikov:
@hchiorean , `RollbackException` might have `null` deferred exception. What about analysis of the cause. If cause is assigned then throw original exception or `SystemFailureException` deferred 'RollbackException` otherwise?

hchiorean:
@okulikov : sounds good; 
With this change the multiple `catch` blocks should be revisited. For example, we can collapse `RollbackException` | `HeuristicRollbackException` | `HeuristicMixedException` into one. 
Also, for these 3 exceptions after catching them and before deciding what exception to re-throw, we should make sure that `txn=null` because in these cases we _never_ want the following to be executed:
```
 catch (Throwable t) {
    // any other exception/error we should rollback
   if (txn != null) {
      txn.rollback();
   }
  // let the exception bubble up
  throw t;
 }
```
since the transaction has already been rolled back.

hchiorean:
looks good to me

rhauch:
@hchiorean, @okulikov: looks good to me.

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/WritableSessionCache.java
```diff
@@ -931,8 +933,8 @@ public void save( Set<NodeKey> toBeSaved,
 
         } catch (RuntimeException e) {
             throw e;
-        } catch (Exception e) {
-            throw new WrappedException(e);
+        } catch (Throwable t) {
```
hchiorean: @okulikov: just saw this: catching `Throwable` here means `RuntimeException` gets wrapped, and we don't want that. So the code should be reverted to `Exception` (same in the other places as well)

hchiorean:
@okulikov : just a minor thing - see above - where a change should be reverted

okulikov:
@hchiorean , I think no since `RuntimeException` is handled separately and at the same level as `Throwable`.

hchiorean:
@okulikov: u'r right; I thought the `catch Throwable` was outside the first set of `catch` blocks.

PR #1596, opened 2016-09-21
MODE-2631: allow for better granularity when configuring JCR lock cleanup process.

dnillia:
The garbage collection process should now have new properties exposed (the property values are provided as an example):
```json
"garbageCollection" : {
    "lockCleanup" : {
        "lockCleanupInitialDelayInMinutes" : 60,
        "lockCleanupIntervalInMinutes" : 180
    }
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,6 @@ ObjectStore
 # Atomikos
 *.lck
 *.epoch
+
+# ignore checkstyle files
+*.checkstyle
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/JcrRepository.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/JcrRepository.java
@@ -1317,13 +1317,17 @@ public class JcrRepository implements org.modeshape.jcr.api.Repository {
                 String threadPoolName = gcConfig.getThreadPoolName();
                 long gcInitialTimeInMillis = determineInitialDelay(gcConfig.getInitialTimeExpression());
                 long gcIntervalInMillis = gcConfig.getIntervalInMillis();
-             
+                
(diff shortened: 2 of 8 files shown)


hchiorean:
@dnillia thanks for submitting this improvement 

I'm not convinced by its usefulness: as mentioned previously, ModeShape already provides a way of configuring garbage collection via
```
"garbageCollection" : {
        "threadPool" : "modeshape-gc-pool",
        "initialTime" : "04:01",
        "intervalInHours" : 5,
    }
```
which includes both locks and binary storage cleanup.

The fact that this runs every X hours is intentional, as a lower time granularity would most likely be detrimental performance wise.

Also, the garbage collection process is really an advanced setting, not something which meant to be used as a "regular setting" for the repository. In other words, users should not rely on the garbage collection process to fix their particular use cases. 

JCR/ModeShape is an API (not a "product") so it's reasonable to expect that depending on the usage context, users should handle exceptional use cases in their application, not push that down to ModeShape. I'm mentioning this because according to [this forum discussion](https://developer.jboss.org/message/963164) you're submitting this in an attempt to fix some custom "atomicity logic" around a series of JCR lock/unlock operations. 

So to conclude, I'm not convinced this should be merged since:
* it adds/expands an already complicated repository configuration file - we already have a huge number of config options
* the GC logic has been present since ModeShape 3 and no one else in the community or otherwise has required this type of advanced setting

@rhauch any thoughts on this ?

rhauch:
@dnillia, @hchiorean I do have some questions about the use case and the need for something like this, so I'll post them on the [forum discussion](https://developer.jboss.org/message/963164).

hchiorean:
closing this, based on [this discussion](https://developer.jboss.org/message/963205#963205)

PR #1597, opened 2016-09-22
MODE-2629 Fixes the issue with too many parameters for the IN clause

hchiorean:
This commit changes the queries that get sent down to the DB so that when the number of parameters exceeds a certain limit, the IN clause gets broken down into multiple `IN OR IN` clauses.

It also adds Docker support for the `modeshape-persistence-relational` module so that low level DB operations can be tested against different DBs.


rhauch:
@hchiorean, looks good.

PR #1598, opened 2016-09-23
MODE-2633 Fixes the handling of lock expiration

hchiorean:
The previous code did no take into account that depending on the timeout setting, a session or an open scoped lock may be expired and still be present on a node.


rhauch:
@hchiorean, looks good.

PR #1599, opened 2016-09-26
MODE-2634 Replaces ConcurrentLinkedHashmap with Caffeine 2.3.3

hchiorean:
The former is dormant and was designed for pre JDK 8 while the latter is the current active project which uses JDK 8

### DIFF
--- a/boms/modeshape-bom-embedded/pom.xml
+++ b/boms/modeshape-bom-embedded/pom.xml
@@ -52,7 +52,7 @@
         <version.org.jgroups>3.6.8.Final</version.org.jgroups>
         <version.com.h2>1.4.191</version.com.h2>
         <version.com.zaxxer.HikariCP>2.4.3</version.com.zaxxer.HikariCP>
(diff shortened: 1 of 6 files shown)


hchiorean:
unfortunately even though the build is passing on my machine, in Travis it seems to be causing deadlocks in `org.modeshape.jcr.JcrWorkspaceTest`

Since there isn't a way to get a thread dump from Travis, I'm not comfortable merging this. We'll leave the JIRA open and at some point later on we can reevaluate this against a newer version of the library perhaps.

hchiorean:
reopened and amended to run the cache's eviction synchronously (based on [this comment](https://issues.jboss.org/browse/MODE-2634?focusedCommentId=13298592&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13298592))

ben-manes:
wonderful, the tests passed! :)

hchiorean:

rhauch:
@hchiorean, looks good.

PR #1605, opened 2016-11-14
Mode-2620-b Method node.getParent() throws NullPointerException

okulikov:
This PR contains following changes:
- Pure JDBC database clean up between persistence integration tests;
- Transaction failure handling causing data lost or corruption;
- Makes creating projection atomic.

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/WritableSessionCache.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/WritableSessionCache.java
@@ -35,10 +35,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
-import javax.transaction.HeuristicMixedException;
-import javax.transaction.HeuristicRollbackException;
-import javax.transaction.NotSupportedException;
-import javax.transaction.RollbackException;
 import javax.transaction.Status;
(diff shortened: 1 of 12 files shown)


modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/WritableSessionCache.java
```diff
@@ -948,6 +869,22 @@ public void save( Set<NodeKey> toBeSaved,
...
+     * 
+     * @param txn the transaction 
+     * @param cause roll back cause
+     * @throws Exception roll back cause.
+     */
+    private void rollback(Transaction txn, Exception cause) throws Exception {
+        try {
+            txn.rollback();
+        } catch (Exception e) {
+        } finally {
```
hchiorean: I think we should at least log this exception here before throwing other one

hchiorean:
looks good, with one minor comment

modeshape-jcr/src/main/java/org/modeshape/jcr/cache/document/WritableSessionCache.java
okulikov: I tried to hide possible exceptions produced by the `cause` and fire only the original exception.

hchiorean: that's fine, but in some cases the fact that we cannot rollback a transaction might provide some useful information as to why that is. That's why logging instead of ignoring this exception might be of more use.

okulikov:
@hchiorean , done.

PR #1606, opened 2016-11-15
MODE-2637 : NullPointerException in IndexChangeAdapter

okulikov:
PR provides fix which substitutes empty array instead of `null` as value for not assigned multiple-value property.

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/spi/index/provider/IndexChangeAdapters.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/spi/index/provider/IndexChangeAdapters.java
@@ -834,9 +834,13 @@ public class IndexChangeAdapters {
                     newValues.addAll(Arrays.asList(newProperty.getValuesAsArray()));
                 }
             } else if (mixinsTypeChange instanceof PropertyAdded) {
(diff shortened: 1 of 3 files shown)


modeshape-jcr/src/main/java/org/modeshape/jcr/spi/index/provider/IndexChangeAdapters.java
```diff
@@ -803,6 +803,11 @@ public NodeTypesChangeAdapter( String property,
             assert this.property != null;
         }
         
+        private static Object[] valuesOf(Property property) {
+            return property.getValuesAsArray() != null ? 
+                    property.getValuesAsArray() : new Object[]{};
```
hchiorean: I think `new Object[0]|` is more concise, since an empty array initializer doesn't make sense

hchiorean:
@okulikov : this fixes the symptom of the bug, but I'm not convinced it fixes the original cause: the current code checks that a property `!isEmpty()` prior to reading the values from the property. However, it seems to me that `isEmpty()` return `false` for a multi-value property with a `null` value ?!

If that is the case, IMO this should be fixed and not the current extra checks from `IndexChangeAdapter`

hchiorean:
@okulikov: looks great

PR #1607, opened 2016-11-24
MODE-2645 : Elasticsearch index provider always recreates index even if they exist upon engine restart

okulikov:
This PR provides the fix which does not delete and the re-create existing index.


hchiorean:
@okulikov this looks good, but I think having a regression test for this in `EsIndexProviderTest` would also be very useful

PR #1608, opened 2016-12-05
MODE-2644 Fixes the reloading of index definitions so that dynamic definitions are activated after a restart

hchiorean:


rhauch:
@hchiorean looks good.

PR #1609, opened 2016-12-06
MODE-2642 Removes the RepositoryFeaturesDetector optimization for #isCheckedOut and #isLocked

hchiorean:
This was introduced as part of https://issues.jboss.org/browse/MODE-2277 to optimize the read access of the `node#isCheckedOut` method, but the entire concept is flawed when user transactions are used. User transactions are isolated from the repository and do not trigger any notifications until the transaction is committed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,9 +27,13 @@ jdk:
   
 # We use docker for the persistence integration tests
 services:
-    - docker
-    
+  - docker
+
+before_script:
+  - sudo docker ps -a 
+  
 # We have to use our local `settings.xml`, and the clustering tests require IPv4 on the Travis infrastructure
 # We're also changing the Maven default logging to avoid printing out all the 'Downloading' messages which can cause the Travis
 # build to fail when the log size exceeds 4MB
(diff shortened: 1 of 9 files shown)


hchiorean:
@rhauch  another `Error starting userland proxy: listen tcp 0.0.0.0:3306: bind: address already in use (Internal Server Error: 500)` failure, even though ModeShape uses only 1 instance of a docker container which should be started/stopped once during the `persistence-tests`

I'm not sure wth is going on, the same build passes fine on my machine. Any ideas ?

rhauch:
@hchiorean try using a different port. I think Travis-CI changed recently and now port 3306 is already bound even though we don't set up Travis' MySQL server.

hchiorean:
@rhauch : the exact same error occurs with Postgres and port 5432. I'm not sure adding custom configuration files for all the docker images is desirable. Actually, it's not even possible for PG since the base docker image exposes 5432 explicitly.

rhauch:
I'll file an issue with Travis-CI, and hopefully they will correct their environment. In the meantime, changing the port for the MySQL container should work.

hchiorean:
@rhauch that did the trick, thanks

hchiorean:
@rhauch are you ok with the changes ?

rhauch:
@hchiorean looks good.

PR #1611, opened 2016-12-09
MODE-2636 Fixes the missing behavior of parsing the table-name attribute for DB persistence on Wildfly

hchiorean:


rhauch:
Looks good.

PR #1614, opened 2016-12-12
MODE-2635 Adds replica-set support for the Mongo binary store to the JSON and WF configs

hchiorean:


hchiorean:
@rhauch are you with merging this ?

rhauch:
@hchiorean Rather than use `replica-sets` as the configuration properties, I think we should use `host-addresses` so that we can reuse it for sharded clusters if we ever need to. We still need to keep the `host` and `port` config properties for backward compatibility.

Then, I would suggest changing the MongodbBinaryStore class to use `List<ServerAddress>` in place of the `host`, `port`, and `replicaSets` fields. The ServerAddress given by the `host` and `port` should just be placed first in the `List<ServerAddress>`, as long as that list contains unique values. This way we can handle all of the configuration properties in the most meaningful and natural way.

hchiorean:
amended the commit

rhauch:
It'd be great if 'host' and 'port' were made optional in the configuration, but I understand why that might be an issue. Looks good.

hchiorean:
I could make them optional both in the JSON schema and WF schema if you feel that's important. That would require some custom validation though to make sure either `host` and `port` or `hostAddresses` are present

rhauch:
Would be nice if we could use only `host-addresses` in newer or updated deployments, but making sure either exists is the hard part. I'll leave it up to you to decide whether the custom validation is worth it.

hchiorean:
I've added another commit which makes `host` and `port` optional from a configuration perspective and performs the validation in the `MongoDBBinaryStore` constructor so that either `host:port` or `hostAddresses` are provided

rhauch:
@hchiorean awesome! Looks great.

PR #1616, opened 2016-12-14
MODE-2641 Changes the behavior of lock#getSecondsRemaining() to return the actual number of seconds for open scoped locks

hchiorean:


hchiorean:

rhauch:
@hchiorean looks good.

PR #1617, opened 2016-12-15
MODE-2643 Changes the GitConnector's logic so that it always returns external binaries

hchiorean:
The previous code attempted to store binaries for large files into the repository's binary store which is not possible outside of a transaction. In general, connectors should always return external binaries which are resolved by the external system "on demand"


rhauch:
Looks good.

PR #1619, opened 2016-12-20
MODE-2638 Adds additional validations that WF clustering configuration is passed down correctly

hchiorean:


rhauch:
Looks good.

PR #1620, opened 2017-01-05
MODE-2654 Fixes the behavior of copying locked nodes

hchiorean:
As per `JCR #17.7` locked nodes should be copied without any special restriction


rhauch:
Looks good.

PR #1621, opened 2017-01-16
MODE-2635 Makes the 'host' and 'port' attributes optional for the MongoDB binary store when 'host-addresses' is used

hchiorean:


rhauch:
Looks good.

PR #1622, opened 2017-01-17
MODE-2658 Fixes the sequencing's behavior of removing existing output when the same node is sequenced multiple times

hchiorean:


rhauch:
Looks good.

PR #1623, opened 2017-01-18
MODE-2659 Updates the JBoss AS artifact versions to use and be compliant with Wildfly 10.1.0.Final

hchiorean:


rhauch:
Looks good.

PR #1624, opened 2017-02-02
MODE-2663 Adds ordering to the MongoDB input/output stream chunks so that they work with Mongo replicas

hchiorean:
based on the patches provided on the [MODE-2663](https://issues.jboss.org/browse/MODE-2663) with slight modifications

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/MongodbBinaryStore.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/MongodbBinaryStore.java
@@ -26,6 +26,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
+import org.modeshape.common.annotation.NotThreadSafe;
 import org.modeshape.common.util.IoUtil;
 import org.modeshape.common.util.StringUtil;
 import org.modeshape.jcr.JcrI18n;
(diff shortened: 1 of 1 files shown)


rhauch:
How does this work with existing databases?

hchiorean:
it shouldn't impact them: it's a new field being added to a collection plus a new index being added.
The whole logic is based on latter generating the equivalent of an `orderBy` query on the former

As far as I understand the Mongo docs, if the new field is not present (i.e. old data) the Mongo treats the field as having `NULL` values in the `orderBy` clause effectively placing them first in the result. So from the docs the first part of the change seems fine.

The second part - running a `sort` on an index which does not exist (i.e. on old data) - I'm not sure what happens. A supplementary `look if the index is present` check  can be added to the code, but that would be done each time a binary value is read.

Since I don't have a lot of experience with Mongo, I can ask the author of the issue to check the behavior in the second case.

hchiorean:
I've added more explicit questions [in this comment](https://issues.jboss.org/browse/MODE-2663?focusedCommentId=13357443&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13357443)

hchiorean:
@rhauch: I've amended the commit and added versioning logic to the chunks.
This should ensure that:
* the logic for reading old data remains unchanged without the overhead of looking up indexes and potential issues with missing fields/index names
* we can enhance this in the future (should be need to) in a backwards-compatible manner

modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/MongodbBinaryStore.java
```diff
@@ -192,8 +199,10 @@ public BinaryValue storeValue( InputStream stream,
 
             // store content
             DBCollection content = db.getCollection(key.toString());
-            ChunkOutputStream dbStream = markAsUnused ? new ChunkOutputStream(content, System.currentTimeMillis()) : new ChunkOutputStream(
-                                                                                                                                           content);
+            content.createIndex(DATA_CHUNK_SORT_INDEX);
```
rhauch: This method is called every time a value is stored. Is that required to add the document to the index? And is it safe to be called repeatedly? 

And IIUC, when this method does build the index it will be [done in the foreground](http://docs.mlab.com/indexing/). Is that correct? Would it be better to be done in the background?

hchiorean: each binary value collection must have its own index, so that's the reason this is called each time
note that `db.getCollection(key.toString());` will create a new collection for each binary value

I don't know the internals of how Mongo manages Indexes on collections, but I would not try doing this in the background because that would mean adding synchronization between creating a new binary value and reading that binary value, in the sense that until the index is created one cannot read the value back. This complexity is simply not worth it IMO.

rhauch: > each binary value collection must have its own index, so that's the reason this is called each time. Note that `db.getCollection(key.toString());` will create a new collection for each binary value

Hmm, okay. Not sure whether that's a good idea, but I'm not familiar with the limitations or constraints on the number of MongoDB collections. Besides, that's what we've got.

When I mentioned "done in the background", I was not referring to us doing this asynchronously. Instead, [this document](http://docs.mlab.com/indexing/#background-vs-foreground-index-builds) says that one can add a `background=true` field to the index to have MongoDB do it asynchronously. But, this probably doesn't matter because the time to create an index on a newly-created collection should be negligible.

rhauch:
This looks good to me. Would be good to have [Patrick](https://issues.jboss.org/browse/MODE-2663) review it as well.

hchiorean:
@rhauch are you ok with the current form of this PR ?

PR #1626, opened 2017-02-08
MODE-2664 Fixes various issues with SQLServer support

hchiorean:
The fixes include:
* making sure CREATE and DROP table statements correctly detect existing objects
* using correct syntax for DB locking
* using the correct binary type for storing schematic documents


dbh:
will try it out. TY

PR #1627, opened 2017-02-10
MODE-2661 Changes the logic of some of the Lucene queries to improve search performance

hchiorean:
The changes include both the logic of `ConstantScoreWeightQuery` which no longer goes through all the documents directly and also making sure to use Lucene native `TermQueries` when applicable. They also refactor some of the query code to use Java 8 idioms.

This commit also upgrades the Lucene version to the latest 6.4.1


hchiorean:

PR #1629, opened 2017-02-15
MODE-2660 Removes modeshape-client and moves the packaging logic for the fat jar to the modeshape-jdbc module

hchiorean:


rhauch:
@hchiorean, just to be clear: this change combines both the remote JDBC driver functionality and the REST API client library into a single fat JAR?

hchiorean:
@rhauch: prior to this change (starting with ModeShape 4), `modeshape-client`'s only function was to create a FAR JAR which contained `modeshape-jdbc` plus transitive dependencies, all of them together representing a self-contained remote JDBC driver (remote = able to connect to repositories via HTTP)

This change effectively moves this packaging activity to `modeshape-jdbc` module where it arguably belongs. In other words, `modeshape-jdbc` will now produce during `assembly` an additional artifact with the `driver` classifier representing the aforementioned JAR.

The original reason for `modeshape-client` was, in addition to the JDBC remote driver, the `IRestClient` interface and some other Designer publishing code. However, as agreed a while back, this has been removed since ModeShape 4.

rhauch:
@hchiorean IIRC, we agreed a while back that we'd forgo this change. If so, can we close this PR without merging?

hchiorean:
Closing without merging.

PR #1631, opened 2017-02-24
MODE-2667, Added S3 endpoint support

open4storage:
Added changes to:

-Added additional overloaded  constructor to take 'endpoint' as an parameter
-Updated schema file to support this additional parameter (as optional to keep backward compatibility).
-Repository config files so it uses endpoint
{
  "name": "Test Repository",
  "storage": {
    "binaryStorage": {
      "type": "s3",
      "username": "access_key",
      "password": "secrey",
      "bucketName": "modeshape_bucket",
      "endPoint": "https://some3scompatiblestorage"
    }
  }
}

### DIFF
--- a/deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/AddS3BinaryStorage.java
+++ b/deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/AddS3BinaryStorage.java
@@ -41,7 +41,8 @@ public class AddS3BinaryStorage extends AbstractAddBinaryStorage {
         binaries.setString(FieldName.USER_PASSWORD, password);
         String bucketName = ModelAttributes.S3_BUCKET_NAME.resolveModelAttribute(context, model).asString();
         binaries.setString(FieldName.BUCKET_NAME, bucketName);
-        String endPoint = ModelAttributes.S3_ENDPOINT_URL.resolveModelAttribute(context, model).asString();
+        ModelNode node = ModelAttributes.S3_ENDPOINT_URL.resolveModelAttribute(context, model);
+        String endPoint = node.isDefined() ? node.asString() : null;  //check if the node exist before getting the value
         binaries.setString(FieldName.ENDPOINT_URL, endPoint);
     }
 
--- a/deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/AddS3BinaryStorage.java
+++ b/deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/AddS3BinaryStorage.java
@@ -41,6 +41,8 @@ public class AddS3BinaryStorage extends AbstractAddBinaryStorage {
         binaries.setString(FieldName.USER_PASSWORD, password);
         String bucketName = ModelAttributes.S3_BUCKET_NAME.resolveModelAttribute(context, model).asString();
         binaries.setString(FieldName.BUCKET_NAME, bucketName);
+        String endPoint = ModelAttributes.S3_ENDPOINT_URL.resolveModelAttribute(context, model).asString();
+        binaries.setString(FieldName.ENDPOINT_URL, endPoint);
     }
 
     @Override
--- a/deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/ModelAttributes.java
+++ b/deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/ModelAttributes.java
@@ -981,6 +981,14 @@ public class ModelAttributes {
             .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
             .build();
 
+    public static final MappedSimpleAttributeDefinition S3_ENDPOINT_URL =
+        new MappedAttributeDefinitionBuilder(Attribute.ENDPOINT_URL.getLocalName(), ModelType.STRING,
+                                             FieldName.STORAGE, FieldName.BINARY_STORAGE, FieldName.ENDPOINT_URL)
+                .setAllowExpression(true)
+                .setAllowNull(true)
+                .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
+                .build();
+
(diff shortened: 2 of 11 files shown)


modeshape-jcr/src/main/java/org/modeshape/jcr/RepositoryConfiguration.java
```diff
@@ -1211,8 +1212,17 @@ public BinaryStore getBinaryStore() throws Exception {
                 String username = binaryStorage.getString(FieldName.USER_NAME);
                 String password = binaryStorage.getString(FieldName.USER_PASSWORD);
                 String bucketName = binaryStorage.getString(FieldName.BUCKET_NAME);
-                store = new S3BinaryStore(username, password, bucketName);
+
+                //Use S3 provided endpoints
+                if (FieldName.ENDPOINT_URL != null) {
```
hchiorean: this `if` doesn't seem right; it's always true :)
Perhaps the intent is to check if the `endPoint` local variable is not null

hchiorean:
@open4storage thanks for contributing the enhancement; there is one small issue with the PR (see comment), but other than that it looks fine to me

modeshape-jcr/src/main/java/org/modeshape/jcr/RepositoryConfiguration.java
hchiorean: or, since the `S3BinaryStore` constructor is `null` aware of the `endPoint` parameter, the entire `if/else` logic isn't required. Once can simply use the new 4 arg constructor always

open4storage: Hi @hchiorean I enabled allow null on the schema, so it branches depending of the use of this optional argument in the repo config and calls the appropriate overloaded constructor.  On my tests this segment is branching properly for both S3 (3 args) and S3 compatible storage (4 args).

deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/ModelAttributes.java
```diff
@@ -981,6 +981,14 @@
             .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
             .build();
 
+    public static final MappedSimpleAttributeDefinition S3_ENDPOINT_URL =
+        new MappedAttributeDefinitionBuilder(Attribute.ENDPOINT_URL.getLocalName(), ModelType.STRING,
+                                             FieldName.STORAGE, FieldName.BINARY_STORAGE, FieldName.ENDPOINT_URL)
+                .setAllowExpression(true)
+                .setAllowNull(true)
```
open4storage: @hchiorean This is where I allowed the null to detect if the optional parameter is there in order to branch to the right constructor.

modeshape-jcr/src/main/java/org/modeshape/jcr/RepositoryConfiguration.java
hchiorean: @open4storage the `if (FieldName.ENDPOINT_URL != null)` doesn't make sense, since the constant is always there.

deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/ModelAttributes.java
hchiorean: allowing `null` is fine and is the right way to ensure backwards compatibility

hchiorean:
@open4storage there's an additional change required to the AS kit which I missed on the first review:
the new attribute should also be added (for documentation purposes) to the [modeshape.xsd](https://github.com/ModeShape/modeshape/blob/master/deploy/jbossas/modeshape-jbossas-subsystem/src/main/resources/schema/modeshape_3_0.xsd#L885) schema and to the [local descriptors](https://github.com/ModeShape/modeshape/blob/master/deploy/jbossas/modeshape-jbossas-subsystem/src/main/resources/org/modeshape/jboss/subsystem/LocalDescriptions.properties#L256)

deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/AddS3BinaryStorage.java
```diff
@@ -41,6 +41,8 @@ protected void writeBinaryStorageConfiguration( String repositoryName,
         binaries.setString(FieldName.USER_PASSWORD, password);
         String bucketName = ModelAttributes.S3_BUCKET_NAME.resolveModelAttribute(context, model).asString();
         binaries.setString(FieldName.BUCKET_NAME, bucketName);
+        String endPoint = ModelAttributes.S3_ENDPOINT_URL.resolveModelAttribute(context, model).asString();
```
hchiorean: this won't work correctly if `ModelAttributes.S3_ENDPOINT_URL` is not defined (the `asString()` part). It will actually return `undefined` instead of `null`.

So the correct code should look like:
```
ModelNode node = resolveModelAttribute();
String endpoint = node.isDefined() ? node.asString() : null
```

modeshape-jcr/src/main/java/org/modeshape/jcr/RepositoryConfiguration.java
open4storage: I see what you meant with the constant... fixing shortly.  Thx!I see what you meant with the constant... fixing shortly.  Thx!

open4storage: I see what you meant with the constant... fixing shortly.  Thx!  I see what you meant with the constant... fixing shortly.  Thx!

deploy/jbossas/modeshape-jbossas-subsystem/src/main/java/org/modeshape/jboss/subsystem/AddS3BinaryStorage.java
open4storage: Good catch... fixing this one too.  Thx!

hchiorean:
@open4storage updated PR looks good; thanks
@rhauch are you ok with this enhancement ?

rhauch:
Looks good.

PR #1633, opened 2017-03-02
MODE-2668, MODE-2669 Fixes some more issues around user transactions

hchiorean:
The change for https://issues.jboss.org/browse/MODE-2668 detects aborted user transactions when the equivalent ModeShape transaction is being searched

The change for https://issues.jboss.org/browse/MODE-2669 allows the relational DB provider to correctly clean up resources even if commit or rollback are called from different threads


hchiorean:
@rhauch are you with this ?

rhauch:
Looks good

PR #1636, opened 2017-03-10
MODE-2670 Changes the logic of initializing in a cluster, removing redundant locking

hchiorean:
This commit also changes the default locking method in a cluster to be DB locking, which is far more reliable than JGroups. That latter has been deprecated (code wise) and will be removed in the next major version.

### DIFF
--- a/integration/modeshape-jbossas-integration-tests/src/test/java/org/modeshape/test/integration/ClusteredConfigurationIntegrationTest.java
+++ b/integration/modeshape-jbossas-integration-tests/src/test/java/org/modeshape/test/integration/ClusteredConfigurationIntegrationTest.java
@@ -77,8 +77,8 @@ public class ClusteredConfigurationIntegrationTest {
     @Test
     @FixFor({"MODE-1923", "MODE-1929", "MODE-2226"})
     public void clusteredRepositoryShouldHaveStartedUpUsingExternalJGroupsConfigFile() throws Exception {
-        checkRepoConfiguration(clusteredRepository1, "modeshape-wf-it1", true, false);
-        checkRepoConfiguration(clusteredRepository2, "modeshape-wf-it1", true, false);
+        checkRepoConfiguration(clusteredRepository1, "modeshape-wf-it1", true, true);
+        checkRepoConfiguration(clusteredRepository2, "modeshape-wf-it1", true, true);
         checkRepoStarted(clusteredRepository1, clusteredRepository2);
     }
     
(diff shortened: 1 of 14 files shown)


hchiorean:

rhauch:
@hchiorean my only concern is how this impacts existing users that will upgrade.  

Because this changes the default (which is sensible and I completely agree with), users that don't explicitly configure the locking method will have different behavior upon upgrade and restart, ergo they need to change their configuration after upgrade if that's not acceptable and they (for some reason) want to keep using the JGroups locking.

So given this, what happens if anything when a user switches from JGroups locking to DB locking? Since the cluster would have to be upgraded and restarted all at once, no locks would be held which means no lock migration is necessary. Might they have to alter their DB schema to add the tables, or will that all be transparent to them? Do they have to do any work other than upgrade and restart?

hchiorean:
>So given this, what happens if anything when a user switches from JGroups locking to DB locking?

locking should be completely transparent from the perspective of repository content - i.e. neither JGroups nor DB locking, store any dedicated repository content. They both rely on the same `REPOSITORY_INFO` entry for initialization, which is unchanged by this commit

>Since the cluster would have to be upgraded and restarted all at once, no locks would be held which means no lock migration is necessary. 

again, these are not JCR locks (i.e. locks which get persisted) but ModeShape's internal mechanism which _should_ ensure strong serializable consistency (coupled with DB transactions). They are not persisted in any way

>Might they have to alter their DB schema to add the tables, or will that all be transparent to them? Do 
they have to do any work other than upgrade and restart?

upgrading doesn't some into play, since no persistence data has changed.

So someone moving from a version prior to 5.4 to 5.4 and keeping their old config, should see absolutely no difference (except that behind the scenes some other logic takes place each time a session tries saving content) 

Even if someone does an incremental upgrade to an existing cluster - essentially mixing old and new code in the same active cluster - as long as the repository was initialized previously - the new logic should have no negative impact since the `REPOSITORY_INFO` is already persisted.

The only *potential* problem would be a cluster of mixed old and new code (pre 5.4 and 5.4) starting up for the 1st time. I really hope no one is attempting this.. (or has ever attempted these type of mix-matches)

A couple of additional mentions: 
* JGroups event messaging (and hence JGroups in general) is still present since that's the only way to notify other cluster nodes of changes (and trigger workspace cache flushing)
* JGroups locking is still possible but is now made explicit - i.e. users can use it has using `locking:jgroups` in their configuration

rhauch:
@hchiorean sounds good. You addressed all of my concerns, and IMO we should not need to handle heterogenous clusters of multiple versions.

PR #1638, opened 2017-03-14
MODE-2674. Modeshape 5 open database connections with HikariCP.

mohideen:
Configure the minimum idle connection property of the HikariDataSource.

https://issues.jboss.org/browse/MODE-2674


hchiorean:
@mohideen it's good enhancement, thanks for submitting the patch

I will do a bit of investigation to see how easy/hard it would be to actually support all the `HikariConfig` params transparently from ModeShape.  I think that would provide the greatest degree of flexibility.

hchiorean:
Closing this, as https://github.com/ModeShape/modeshape/pull/1641 adds generic support for custom Hikari configuration options.
The default `minimumIdle` will be set to 1.

PR #1641, opened 2017-03-17
MODE-2674 Adds the ability to pass custom Hikari configuration options to the relational db provider

hchiorean:
It configures the `minimumIdle` property to a default of `1`, meaning only 1 idle connection should be in the pool at any time.  It  changes the structure of the WF kit by removing the custom Hikari module (which had just 1 resource) and moving that dependency inside the persistence-relational module.

It also updates the Hikari version to `2.6.1`.


hchiorean:
unfortunately this change seems incomplete, as all sorts of weird persistence integration test failures start happening.

PR #1642, opened 2017-03-20
MODE-2675 Fixes the parsing and extraction of WebDAV lock tokens from request headers

hchiorean:


jacobilsoe:
Do you plan to also take the ETag into account? It looks like it is being ignored in the pull request.

web/modeshape-webdav/src/main/java/org/modeshape/webdav/methods/AbstractMethod.java
```diff
@@ -116,32 +123,11 @@ protected AbstractMethod() {
...
 
     public static String lastModifiedDateFormat( final Date date ) {
-        DateFormat df = thLastmodifiedDateFormat.get();
-        if (df == null) {
-            df = new SimpleDateFormat(LAST_MODIFIED_DATE_FORMAT, Locale.US);
-            df.setTimeZone(TimeZone.getTimeZone("GMT"));
-            thLastmodifiedDateFormat.set(df);
-        }
-        return df.format(date);
+        return LAST_MODIFIED_DATE_FORMAT.format(date);
```
rhauch: `SimpleDateFormat` is not thread safe, so you can't share it across threads. The static instance _can_ be used across threads since this method is static, so it's not safe to make this change.

web/modeshape-webdav/src/main/java/org/modeshape/webdav/methods/AbstractMethod.java
```diff
@@ -116,32 +123,11 @@ protected AbstractMethod() {
...
-    }
-
-    public static String getLocalDateFormat( final Date date,
-                                             final Locale loc ) {
-        DateFormat df = thLocalDateFormat.get();
-        if (df == null) {
-            df = new SimpleDateFormat(LOCAL_DATE_FORMAT, loc);
-        }
-        return df.format(date);
+        return CREATION_DATE_FORMAT.format(date);
```
rhauch: `SimpleDateFormat` is not thread safe, so you can't share it across threads. The static instance _can_ be used across threads since this method is static, so it's not safe to make this change.

hchiorean:
@jacobilsoe: no, this PR only fixes the fact that lock token were not correctly parsed. The WebDAV support does not handle ETags as part of lock tokens (it never did)

If you want some sort of special handling, we're happy to accept any enhancement requests & contributions ;)

hchiorean:
@rhauch fixed the mentioned issue (forgot that JDK 8 fixed this with separate classes)

rhauch:
@hchiorean any response for @jacobilsoe's question about ETag?

hchiorean:
@rhauch I replied in a previous comment...

rhauch:
@hchiorean ah, missed that. Everything looks good to me.

PR #1644, opened 2017-03-24
MODE-2672 Adds exception handling for the case when multiple cluster nodes attempt to concurrently lock a node

hchiorean:


hchiorean:
@rhauch  are you ok with this ?

modeshape-jcr/src/main/java/org/modeshape/jcr/locking/DbLockingService.java
```diff
@@ -47,13 +45,13 @@ public DbLockingService(long lockTimeoutMillis, Lockable db) {
...
-        boolean result = false;
+        long start = System.currentTimeMillis();
+        boolean result;
         long timeInMills = TimeUnit.MILLISECONDS.convert(time, unit);
         while (!(result = db.lockForWriting(names)) && 
-               TimeUnit.MILLISECONDS.convert((System.nanoTime() - start), TimeUnit.NANOSECONDS) < timeInMills) {
+               System.currentTimeMillis() - start <= timeInMills) {
             //wait a bit (between 50 and 300 ms)
-            long sleepDurationMillis = 50 + RANDOM.nextInt(251);
+            long sleepDurationMillis = 50 + ThreadLocalRandom.current().nextInt(251);
```
rhauch: Hmm... didn't know about this one.

hchiorean:
I only found out recently about TLR and it's supposed to be far more performant than the traditional `Random`, especially under multi-threaded use (which the old static field was a subject of)

PR #1645, opened 2017-03-29
MODE-2678 Changes the DatabaseBinaryStore to use  transactions instead of auto-committed connections

hchiorean:
It also fixes the case of multiple cluster nodes inserting the same binary simultaneously in the database.

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/DatabaseBinaryStore.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/DatabaseBinaryStore.java
@@ -158,11 +158,12 @@ public class DatabaseBinaryStore extends AbstractBinaryStore {
     public BinaryValue storeValue( InputStream stream, final boolean markAsUnused ) throws BinaryStoreException {
         // store into temporary file system store and get SHA-1
         final BinaryValue temp = cache.storeValue(stream, markAsUnused);
+        // prepare new binary key based on SHA-1
+        final BinaryKey key = new BinaryKey(temp.getKey().toString());
         try {
             return dbCall(connection -> {
-                // prepare new binary key based on SHA-1
-                BinaryKey key = new BinaryKey(temp.getKey().toString());
-            
+                connection.setAutoCommit(false);
+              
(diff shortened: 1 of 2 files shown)


modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/DatabaseBinaryStore.java
```diff
@@ -173,14 +174,29 @@ public BinaryValue storeValue( InputStream stream, final boolean markAsUnused )
...
+                // so try reading again
+                 return dbCall(connection -> {
+                    if (database.contentExists(key, !markAsUnused, connection)) {
+                        return new StoredBinaryValue(DatabaseBinaryStore.this, key, temp.getSize());        
+                    }
+                    // nothing there, so rethrow the original exception
+                    throw e;
+                });
+            }
+            throw e;
```
rhauch: Wouldn't it be better to have a separate catch block for SqlException?

hchiorean: If you look at the [dbCall](https://github.com/ModeShape/modeshape/pull/1645/files#diff-d1dc99bb1634087ce64bcb5a66ecc4e4L329) method, that's the only place where the exception can come from since that's where the transaction would be committed.

In other words, the only way to get a `SQLException` in this method (storeValue) would be to change the signature of `dbCall` which is used by several other places that rely only on `BinaryStoreException`

Either that or have a separate `dbCall` method which throws this.

It's definitively possible (both cases) but both would require more code changes than the proposed change here.

hchiorean: @rhauch if you prefer I can look at either of the above 2 options, although I'm not 100% convinced the changes would be significant

rhauch: Sorry, I missed the `getCause()` call, which makes my suggestion irrelevant. Thanks.

PR #1650, opened 2017-04-12
MODE-2684 Removes the compile time dependency of modeshape-core towards Apache Tika

hchiorean:
The mime type extraction functionality will still function as-is if Tika is present, but now there is also an independent extension-based default which will be used if Tika is not present in the CP at runtime

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/mimetype/MimeTypeDetectors.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/mimetype/MimeTypeDetectors.java
@@ -1,0 +1,81 @@
+/*
+ * ModeShape (http://www.modeshape.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.modeshape.jcr.mimetype;
+
+import org.modeshape.common.logging.Logger;
+import org.modeshape.jcr.Environment;
+import org.modeshape.jcr.RepositoryConfiguration;
+import org.modeshape.jcr.mimetype.tika.TikaContentDetector;
+import org.modeshape.jcr.mimetype.tika.TikaNameOnlyDetector;
+
+/**
+ * Class which decides which {@link MimeTypeDetector} implementation to use at runtime
+ * 
+ * @author Horia Chiorean (hchiorea@redhat.com)
+ */
+public final class MimeTypeDetectors {
+    
+    private static final Logger LOGGER = Logger.getLogger(MimeTypeDetector.class);
+    private static final boolean TIKA_AVAILABLE; 
+    
+    static {
+        boolean tikaAvailable = true;
+        ClassLoader classLoader = MimeTypeDetector.class.getClassLoader();
+        try {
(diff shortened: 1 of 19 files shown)


hchiorean:

modeshape-jcr/src/main/java/org/modeshape/jcr/mimetype/MimeTypeDetectors.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Class which decides which {@link MimeTypeDetector} implementation to use at runtime
+ * 
+ * @author Horia Chiorean (hchiorea@redhat.com)
+ */
+public final class MimeTypeDetectors {
+    
+    private static final Logger LOGGER = Logger.getLogger(MimeTypeDetector.class);
+    private static boolean TIKA_AVAILABLE;
```
rhauch: Couldn't we make this field `final`?

modeshape-jcr/src/main/java/org/modeshape/jcr/mimetype/MimeTypeDetectors.java
```diff
@@ -0,0 +1,80 @@
...
+    
+    private static final Logger LOGGER = Logger.getLogger(MimeTypeDetector.class);
+    private static boolean TIKA_AVAILABLE; 
+    
+    static {
+        TIKA_AVAILABLE = true;
+        ClassLoader classLoader = MimeTypeDetector.class.getClassLoader();
+        try {
+            Class.forName("org.apache.tika.detect.DefaultDetector", true, classLoader);
+            Class.forName("org.apache.tika.metadata.Metadata", true, classLoader);
```
rhauch: Just to be clear, we require both classes to be present? (Sorry, it's been a long time since I've looked at the Tika stuff.)

modeshape-jcr/src/main/java/org/modeshape/jcr/mimetype/tika/TikaMimeTypeDetector.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.modeshape.jcr.mimetype;
+package org.modeshape.jcr.mimetype.tika;
```
rhauch: Are you sure there are no backward compatibility issues or required configuration changes with this moving of the classes?

modeshape-jcr/src/main/java/org/modeshape/jcr/mimetype/MimeTypeDetectors.java
hchiorean: both classes are part of the same jar - `tika-core` and both are used by the ModeShape-Tika integration for mime type detection.

Making sure both are present is therefore an attempt to make sure that a "correct version" of Tika in the classpath

modeshape-jcr/src/main/java/org/modeshape/jcr/mimetype/tika/TikaMimeTypeDetector.java
hchiorean: prior to this change, the Tika mime type detection classes were instantiated explicitly by ModeShape based on the given [repository configuration](https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/main/java/org/modeshape/jcr/RepositoryConfiguration.java#L1243)

In other words, these were not exposed on the outside, essentially making them an internal implementation detail. 

Since the repository configuration has not changed in any way, there should be no backward-compatibility concerns because of the package and name change for these classes.

rhauch: Great to hear. Thanks.

modeshape-jcr/src/main/java/org/modeshape/jcr/mimetype/MimeTypeDetectors.java
rhauch: Ok.

hchiorean:
amended the commit, making the `TIKA_AVAILABLE` member final, via a local variable alias

PR #1652, opened 2017-04-21
MODE-2687: DB2 Support

Waize:
(cherry picked from commit 3dc6b28)

### DIFF
--- a/persistence/modeshape-persistence-relational/src/main/java/org/modeshape/persistence/relational/DB2Statements.java
+++ b/persistence/modeshape-persistence-relational/src/main/java/org/modeshape/persistence/relational/DB2Statements.java
@@ -1,0 +1,62 @@
+/*
+ * ModeShape (http://www.modeshape.org)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


hchiorean:
the changes look fine, but we still need at least a profile which allows DB2 testing (see [this profile](https://github.com/ModeShape/modeshape/blob/master/modeshape-parent/pom.xml#L315) for MySQL)

That also means we need DB2 Maven artifacts for the JDBC drivers and support for the DB2 docker image (https://hub.docker.com/r/ibmcom/db2express-c/) in order to run our tests (see https://github.com/ModeShape/modeshape/blob/master/modeshape-parent/pom.xml#L586)

Waize:
I think we can provide that profile for you.
Unfortunately I don't have any expirience with docker so I would please you do the docker image think later on, okay?

Waize:
My colleague told me that the DB2 JDBC driver is not as free accessable as it is for oracle or postgres for example. You can access it after installing a DB2 database somewhere thats why I cannot provide any artifacts. 
That said it's quit difficult for me to create a DB2 profile for your tests because that will be just copy and paste without any know how for docker etc. 
I can't do any further things here, sorry

hchiorean:
@rhauch any thoughts on this ?
the change seems minimal and solid in terms of having DB2 support
The only downside is the fact that without JDBC Maven drivers, we can't really have testing profiles for this.

rhauch:
Yeah, I'm okay with it given the lack of tests.

hchiorean:
merged in `master`;
@Waize thanks for the contribution

PR #1654, opened 2017-04-27
[MODE-2690] Set content length in ObjectMetadata for S3 uploads

dalbani:
https://issues.jboss.org/browse/MODE-2690


hchiorean:
merged in `master`; thanks @dalbani

PR #1656, opened 2017-04-27
[MODE-2692] Return null MIME type in case of "application/octet-stream"

dalbani:
https://issues.jboss.org/browse/MODE-2692

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/S3BinaryStore.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/S3BinaryStore.java
@@ -68,6 +68,11 @@ public class S3BinaryStore extends AbstractBinaryStore {
     protected static final String EXTRACTED_TEXT_KEY = "extracted-text";
 
     /*
+     * Key for storing boolean which describes if a MIME type has been explicitly set
+     */
+    protected static final String USER_MIME_TYPE_KEY = "user-mime-type";
+
+    /*
      * Key for storing boolean which describes if object is unused
      */
     protected static final String UNUSED_KEY = "unused";
@@ -140,7 +145,11 @@ public class S3BinaryStore extends AbstractBinaryStore {
         try {
             String key = binaryValue.getKey().toString();
             ObjectMetadata metadata = s3Client.getObjectMetadata(bucketName, key);
-            return metadata.getContentType();
+            if (String.valueOf(true).equals(metadata.getUserMetadata().get(USER_MIME_TYPE_KEY))) {
+                return metadata.getContentType();
+            } else {
+                return null;
+            }
         } catch (AmazonClientException e) {
             throw new BinaryStoreException(e);
         }
@@ -152,6 +161,7 @@ public class S3BinaryStore extends AbstractBinaryStore {
             String key = binaryValue.getKey().toString();
             ObjectMetadata metadata = s3Client.getObjectMetadata(bucketName, key);
             metadata.setContentType(mimeType);
+            metadata.addUserMetadata(USER_MIME_TYPE_KEY, String.valueOf(true));
 
             // Update the object in place
             CopyObjectRequest copyRequest = new CopyObjectRequest(bucketName, key, bucketName, key);
--- a/modeshape-jcr/src/test/java/org/modeshape/jcr/value/binary/S3BinaryStoreTest.java
+++ b/modeshape-jcr/src/test/java/org/modeshape/jcr/value/binary/S3BinaryStoreTest.java
@@ -87,6 +87,7 @@ public class S3BinaryStoreTest extends EasyMockSupport {
     public void testGetStoredMimeType() throws BinaryStoreException {
         ObjectMetadata objMeta = new ObjectMetadata();
         objMeta.setContentType(TEST_MIME);
+        objMeta.addUserMetadata(S3BinaryStore.USER_MIME_TYPE_KEY, String.valueOf(true));
         expect(s3Client.getObjectMetadata(BUCKET, TEST_KEY)).andReturn(objMeta);
 
         replayAll();


hchiorean:
I'm not sure I agree with this change (perhaps that related to the forum discussion): if there is metadata stored in S3 which represents a particular mime-type, then that value should be used as is.

In other words, consider the scenario when a user uploads a `*.bin` file to S3: if mimetype detection is enabled, then the correct mime-type should be `application/octet-stream` which is then stored in S3.

With this change, when attempting to read back the mime-type, the binary store would always return `null`, forcing a mime-type re-discovery. 

The S3 binary store should return  a `null` mime-type only if S3 metadata is not stored(present) at all. Otherwise it should return whatever is stored.

hchiorean:
@dalbani if `getContentType` returns `application/octet-stream` even when the mimetype has not been detected yet (i.e. there is no mimetype) then IMO to correctly fix this we need a user-property or something stored in the metadata which indicates if mimetype discovery has taken place or not.

In other words, when the `storeMimeType` we can set this custom user property and then the `getStoredMimeType` would return `null` (like the proposed change) only if the user property has not been set

dalbani:
@hchiorean  
I have changed the implementation (and the commit message accordingly).

hchiorean:
with the latest change the existing unit test is failing:
```
Tests run: 13, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.295 sec <<< FAILURE! - in org.modeshape.jcr.value.binary.S3BinaryStoreTest

testGetStoredMimeType(org.modeshape.jcr.value.binary.S3BinaryStoreTest)  Time elapsed: 0.013 sec  <<< FAILURE!
```

dalbani:
@hchiorean  
By the way, although it might be a bit silly to ask the question just _now_, but: why do binary stores actually care about MIME types??  
I mean, if I add a node and explicitly set a value for the `jcr:mimeType` property, it doesn't get stored via the binary store — but via the standard persistence layer I suppose?
On the other hand, if I let ModeShape detect the MIME type, then the information ends up in the binary store (e.g. a header in S3) — and in the persistence layer as well, right?
I don't understand the logic here... What component is actually the user of the MIME-type-in-binary-store piece of information?

dalbani:
Hum, I stand corrected for my assertion that "if I add a node and explicitly set a value for the `jcr:mimeType` property, it doesn't get stored via the binary store". This is not true.  
But that still doesn't answer my initial question.

Furthermore, if upload the _same_ document to the repository (same SHA1 thus), but this time explicitly with a _different_ MIME type, what's in the binary store doesn't reflect that?!

hchiorean:
@dalbani I'm not sure I understand your question, most notably the `caring` part :)

If you mean to ask why is there mime type detection at all, it's because it was a feature (outside of the JCR spec) since the early versions of Modeshape (I think going back as ModeShape 2). It is used for [sequencing](https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/main/java/org/modeshape/jcr/JcrSession.java#L2013) and [nt:resource](https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/main/java/org/modeshape/jcr/JcrSession.java#L2208) node types.

If you mean to ask why the mime types `store` and `get` methods are on the `AbstractBinaryStore` SPI it is because the semantics of a mime-type belong to a binary value and not to a JCR node. In other words, you can store mime-type information about a binary completely independent of any JCR semantics (or nodes). You can also ask a binary about its mime-type outside of any JCR node. When I say `ask` I mean [ModeShape's binary API](https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr-api/src/main/java/org/modeshape/jcr/api/Binary.java#L62) (note it's really an API not SPI and part of the `-api` module)

This means that 
a) mime-type detection is optional - you can very well have a binary store implementation which does not store & load any mime type information 
b) if you're not using `nt:resource` types or sequencers no automatic mimetype detection takes place

dalbani:
@hchiorean  
OK, thanks for the explanation of the overall picture.
Yet I still find a bit strange that the MIME type information is duplicated in both the `jcr:mimeType` property and the binary store backend (e.g. S3 header).  
I keep thinking on this subject because I see each request to S3 (`get/setStoredMimeType()`) as one additional performance hit (and cost), whereas the data in the header won't be used afterwards.

Speaking of S3, it has occurred to me that the Java SDK seems to handle default MIME type a bit differently than other implementations: https://github.com/aws/aws-sdk-java/issues/1143
In summary, the Java SDK uses `application/octet-stream` as default, whereas the "true" default value in S3 is `binary/octet-stream`.
I was wondering if we could use this latest value as a marker that no MIME type has been set by the user.

dalbani:
And if I may add, this line doesn't seem to have any effect: https://github.com/dalbani/modeshape/blob/master/modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/S3BinaryStore.java#L234
```java
metadata.setContentType(fileSystemCache.getMimeType(cachedFile, key.toString()));
```
My understanding is that `fileSystemCache.getMimeType()` always returns `null`.

hchiorean:
@dalbani the S3 binary store was contributed by the community so it's probably best to continue any S3-related discussions with the original contributor.
I tried to explain the semantics and design decisions behind mime-types in ModeShape, in general.

dalbani:
Hey @bbranan, following the discussion above, could you tell me what was your intention when you wrote:
```java
metadata.setContentType(fileSystemCache.getMimeType(cachedFile, key.toString()));
```

Did you intend to have MIME type detection happen within the local `FileSystemBinaryStore`?

bbranan:
@dalbani The purpose of that line was to support anyone who wanted to call setMimeTypeDetector(), (which is defined in the AbstractBinaryStore) on the S3BinaryStore. If a detector is set with that call, the fileSystemCache,getMimeType() call should use the detector to determine the Mimetype of the file stored on the file system, then include that Mimetype in the call to S3. At least, that's what I had in mind. Given that no detector is set currently, this isn't happening, but the potential is there.

dalbani:
@bbranan Thanks for your feedback, I see what was the original goal.
That gives me input for a functionality I'm currently working on: taking advantage of the local `FileSystemBinaryStore` in order to build a cache on disk, given that objects uploaded in S3 by ModeShape are immutable (except metadata).
I'll open a JIRA when I have a working proof of concept.

PR #1658, opened 2017-04-28
[MODE-2694] Close input streams passed to TikaNameOnlyDetector.detect()

dalbani:
https://issues.jboss.org/browse/MODE-2694

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/mimetype/tika/TikaNameOnlyDetector.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/mimetype/tika/TikaNameOnlyDetector.java
@@ -15,7 +15,6 @@
  */
 package org.modeshape.jcr.mimetype.tika;
 
-import java.io.InputStream;
 import org.apache.tika.detect.Detector;
 import org.apache.tika.metadata.Metadata;
 import org.apache.tika.mime.MediaType;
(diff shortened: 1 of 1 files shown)


modeshape-jcr/src/main/java/org/modeshape/jcr/mimetype/tika/TikaNameOnlyDetector.java
```diff
@@ -61,7 +63,8 @@ protected void initDetector( ClassLoader loader ) {
     @Override
     protected String detect( InputStream inputStream, Metadata metadata ) {
         try {
-            // we never care about the stream here
+            // we never care about the stream here, but close it anyway
+            IoUtil.closeQuietly(inputStream);
```
hchiorean: I would move this inside a `finally` block in the same method

dalbani: Yep, agree.

dalbani: I happen to get this kind of S3-related warnings in relation to this unused stream:
```Not all bytes were read from the S3ObjectInputStream, aborting HTTP connection. This is likely an error and may result in sub-optimal behavior. Request only the bytes you need via a ranged GET or drain the input stream after use.```

Comes from https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/S3AbortableInputStream.java#L159.

hchiorean:
@dalbani this looks ok, with the exception of the previous comment - i.e. moving the logic of closing the stream inside a `finally` block

dalbani:
Moved into `finally` block.

PR #1659, opened 2017-05-01
[MODE-2697] Remove escaping of single quote characters in JSON documents

dalbani:
JSON grammar doesn't allow escaping of single quote characters:

image

### DIFF
--- a/modeshape-schematic/src/main/java/org/modeshape/schematic/internal/document/CompactJsonWriter.java
+++ b/modeshape-schematic/src/main/java/org/modeshape/schematic/internal/document/CompactJsonWriter.java
@@ -170,8 +170,7 @@ public class CompactJsonWriter implements JsonWriter {
         for (char c = iter.first(); c != CharacterIterator.DONE; c = iter.next()) {
             switch (c) {
                 case '"':
-                case '\'':
-                    // Escape the single- or double-quote characters ..
+                    // Escape double-quote characters ..
                     writer.append('\\').append(c);
                     break;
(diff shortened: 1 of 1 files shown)


hchiorean:
Note that ModeShape's JSON writer was never intended to be an exact implementation of the JSON strict spec, but rather a model of Mongon's 2.x BSON writer extension.

The whole single quote escaping is an interesting topic (see http://stackoverflow.com/questions/2275359/jquery-single-quote-in-json-response) but the problem with this change is *it is not backwards compatible*. I would argue towards trying to use the Unicode escape chars for single quotes as a possible workaround.

After this change ModeShape would produce different JSON documents for nodes and as such, this is not something which I  would consider merging.

hchiorean:
@dalbani: I would only consider this change if you were able to produce a test case showing a JCR related bug due to the way the internal documents (nodes) are stored.

Otherwise, changing the `CompactJsonWriter` is not an option for the reason mentioned previously. This is an internal ModeShape class used for creating the JSON-like documents representing JCR nodes.

dalbani:
@hchiorean: then maybe the best way is to switch from using ModeShape's JSON implementation to another ElasticSearch Java client?

hchiorean:
@dalbani: the ES index provider is an optional piece of kit. As such, any problems with it should be fixed in such a way that they do not affect the internal ModeShape storage in a non-backward compatible fashion.

hchiorean:
closing without merging as per the previous discussion.

PR #1660, opened 2017-05-06
[MODE-2698] Improve performance of garbage collection for S3 binary storage

dalbani:
https://issues.jboss.org/browse/MODE-2698

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/S3BinaryStore.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/S3BinaryStore.java
@@ -23,11 +23,14 @@ import com.amazonaws.internal.StaticCredentialsProvider;
 import com.amazonaws.services.s3.AmazonS3Client;
 import com.amazonaws.services.s3.iterable.S3Objects;
 import com.amazonaws.services.s3.model.CopyObjectRequest;
+import com.amazonaws.services.s3.model.GetObjectTaggingRequest;
+import com.amazonaws.services.s3.model.GetObjectTaggingResult;
 import com.amazonaws.services.s3.model.ObjectMetadata;
+import com.amazonaws.services.s3.model.ObjectTagging;
 import com.amazonaws.services.s3.model.S3Object;
 import com.amazonaws.services.s3.model.S3ObjectSummary;
-import org.modeshape.common.logging.Logger;
-import org.modeshape.jcr.JcrI18n;
+import com.amazonaws.services.s3.model.SetObjectTaggingRequest;
+import com.amazonaws.services.s3.model.Tag;
 import org.modeshape.jcr.value.BinaryKey;
 import org.modeshape.jcr.value.BinaryValue;
 
@@ -35,10 +38,11 @@ import javax.jcr.RepositoryException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Collections;
-import java.util.Date;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 /**
  * Binary storage option which manages the storage of files to Amazon S3
@@ -73,9 +77,9 @@ public class S3BinaryStore extends AbstractBinaryStore {
     protected static final String USER_MIME_TYPE_KEY = "user-mime-type";
 
     /*
-     * Key for storing boolean which describes if object is unused
+     * Tag key for storing boolean which describes if object is unused
      */
-    protected static final String UNUSED_KEY = "unused";
+    protected static final String UNUSED_TAG_KEY = "modeshape.unused";
 
     /**
      * Creates a binary store with a connection to Amazon S3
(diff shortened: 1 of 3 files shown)


dalbani:
I don't really understand why TravaisCI fails in part of the codebase which I didn't touch?!

modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/S3BinaryStore.java
```diff
@@ -268,45 +269,61 @@ public InputStream getInputStream(BinaryKey key) throws BinaryStoreException {
...
+        Map<String, String> mergedTags =  initialTags.stream().collect(Collectors.toMap(Tag::getKey, Tag::getValue));
+        mergedTags.put(changeTag.getKey(), changeTag.getValue());
+        return mergedTags.entrySet().stream().map(
+                entry -> new Tag(entry.getKey(), entry.getValue())).collect(Collectors.toList());
+    }
+
+    @Override
+    public void removeValuesUnusedLongerThan(long minimumAge, TimeUnit timeUnit) throws BinaryStoreException {
+        // XXX
+        // What should this method do?
```
dalbani: I was thinking, maybe the original functionality and code should be left in place, as default behavior.
This would be disabled and turned into a no-op if the user explicitly configures the S3 binary store with an option like, say, `out-of-band-deletion = true`.

mbenson: @dalbani can you help to educate the community of nascent ModeShape archaeologists ;) on the old code vs. this patch and the implications of your suggestion here? @bbranan any input from you as the original author of the S3 binary store would also be appreciated.

mbenson: I find that reading MODE-2698 adds a lot of context. For my part, since it appears that using the Amazon lifecycle would require additional configuration on the part of the user, I agree it would be best to leave the original code in place and implement this as a (recommended) optional feature.

modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/S3BinaryStore.java
```diff
@@ -73,9 +77,9 @@
     protected static final String USER_MIME_TYPE_KEY = "user-mime-type";
 
     /*
-     * Key for storing boolean which describes if object is unused
+     * Tag key for storing boolean which describes if object is unused
      */
-    protected static final String UNUSED_KEY = "unused";
+    protected static final String UNUSED_TAG_KEY = "modeshape.unused";
```
mbenson: Wouldn't changing this value break backward compatibility in some way for existing users of the S3 store? Is there a straightforward migration mechanism?

modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/S3BinaryStore.java
bbranan: This approach has the benefit of essentially removing the need for Modeshape to directly manage the removal of any unused files. If a bucket is large, this could provide a significant performance benefit. What is lost is control over purge timing, which could be a significant problem for some use cases. Regardless, this method will almost certainly no longer abide by the contract defined in the Interface (i.e. it will no longer remove any unused values).

I believe there is significant value in this new approach, but sticking with the original behavior as the default seems like the safest model to me, as it is closest to the expected behavior. 

I don't see where the bucket lifecycle rule is put in place. I assume that is yet to be added.

mbenson: My S3-fu is as yet nonexistent. Do APIs exist whereby ModeShape could configure that rule rather than requiring the user to install it?

bbranan: Yes, there are API calls in S3 that would allow for creating the bucket rule, so that could be done within Modeshape. The user would only need to make sure that the permissions (IAM) policy associated with the AWS credentials they supply are sufficient to allow bucket rules to be added.

mbenson: @bbranan Bill, thanks for your response. It appears that @dalbani has lost interest; could you please weigh in on my question at https://github.com/ModeShape/modeshape/pull/1660/files#r233589792 ?

dalbani: If I remember correctly (boy, a year and a half ago already), my idea was to accomplish 2 things regarding the S3 storage backend:
1. use tags instead of copy objects around (due to object immutability) to keep costs down
2. use S3 lifecycle API which happens to support tags, in order to free ModeShape from a possibly tedious task in case of a repository with lots of unused objects.

To be honest, I've moved off S3 quite quickly and hosted my data in Google Cloud Storage. I have written my own GCS-based storage implementation for ModeShape, largely inspired by the S3 one.
I found GCS more appropriate for my needs (no eventual consistency, object properties can be updated, etc).

I happen to be in the process of migrating a ModeShape repository with 3+ TB / 7+ million binaries to Apache Oak, which seems to be the only active JCR implementation.
ModeShape has been quite stable over the last year and a half, although I ran into some limitations with querying in particular (e.g. JOINs) and indexes (take ages to rebuild).
But, yeah, ModeShape is unfortunately dead, even though I never saw any official announcement by the way, right?
Oh, I also wrote a PostgreSQL module to be able to store BSON data from ModeShape as native JSON binary in the database.
The data is not a "black box" anymore and it can be nicely compressed by PostgreSQL!

mbenson: @blafond has taken over as the lead of ModeShape, and I have been trying to assist him in getting the outstanding PRs merged to `master`. This is the last one (modulo one which I am considering as a duplicate). It is my understanding that we are going to try to make a 5.5 release soon, and to keep the project alive going forward. Any information or code you can provide to improve ModeShape, even as you depart :| , would be, as far as I'm concerned, appreciated.

Matt

mbenson:
https://github.com/mbenson/modeshape/tree/pr1660 contains my work in progress, for the interested

modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/S3BinaryStore.java
mbenson: After further understanding the differences between the current code and your work, existing backend storage would break, but not in the way I was thinking. Currently unused values are marked metadata, while you have tried here to mark them using tags, presumably because tags can be filtered by lifecycle rules. So it's not that you've changed the name of the tag, as I originally thought, but that you've changed the marking mechanism. I think we could handle that with an upgrade operation.

PR #1661, opened 2017-06-17
Add try-with-resources statements to ensure streams are closed

vpop:


hchiorean:
looks good to me;
note that there are other places in this class (e.g. the `copy` method) which would benefit from the same change

PR #1663, opened 2017-07-07
Support importing with a binary store hint

weebl2000:
MODE-2707


weebl2000:
Build failed with Invalid maximum heap size: -Xmx=1524m

Should be -Xmx1524m

weebl2000:
@hchiorean Could you look at this?

hchiorean:
fixed upstream

modeshape-jcr/src/main/java/org/modeshape/jcr/JcrValueFactoryWithHint.java
```diff
@@ -0,0 +1,38 @@
...
+package org.modeshape.jcr;
+
+import java.io.InputStream;
+
+import org.modeshape.jcr.value.BinaryValue;
+
+/**
+ * @author Wessel Nieboer
+ */
+public class JcrValueFactoryWithHint extends JcrValueFactory {
```
mbenson: Should there be explicit documentation to the effect that the configured hint is the *default* hint and won't be used for calls to `#createBinary(InputStream, String)`?

weebl2000: Sounds like a good addition. Will implement

weebl2000: Added javadoc.

PR #1666, opened 2017-10-10
correction to send size with S3 request

albertOPT:
when testing I was initially unable to send files to S3 without this additional change.  I figured I should contribute back to the community as this was helpful for my project.


albertOPT:
true, I believe 234 may have undone the initial setting of content length.  I remember at the time getting error from S3 about missing content length before the change, we have been a build from this branch since the change was made.  you're correct it would be cleaner to have removed 232 and moved down to 235 instead of duplicate lines.

mbenson:
The order of calls doesn't concern me either way, but reviewing the code of `ObjectMetadata` I don't see any reason to suppose that `#setContentType()` and `#setContentLength()` would impact one another in any fashion. Any chance you could double check that the relevant functionality is still broken for you using latest `master`?

mbenson:
I'm pretty sure this report must have been a duplicate of #1654 and am recommending it be closed.

PR #1667, opened 2017-12-10
MODE 2720 - Added shouldSearchForLikeConstraintContainingBackslash test

zcc39r:


mbenson:
This test is intended to reproduce the reported bug, correct?

zcc39r:
Sure, this test reproduces the bug.

PR #1668, opened 2018-01-08
Fix parallel uploading in Amazon S3

coraxs:
Binary value must be marked as "used" (unused=false) when storing in fileSystemCache (no matter what is value of "markAsUnused"). In following case it can be crucial:

If couple of threads are trying to add different files in same S3BinaryStore at same time, they will store files in the same fileSystemCache, and when some thread finish, it will delete all files marked as unused from cache (that will be all files stored in cache from another threads), and exception: 

org.modeshape.jcr.value.binary.BinaryStoreException: Unable to find binary value with key "78ed7457b289ba42ff67256f33a88b5bf0ca3012" within binary store at "/tmp/modeshape-binary-store"
at org.modeshape.jcr.value.binary.S3BinaryStore.storeValue(S3BinaryStore.java:251)
at org.modeshape.jcr.value.binary.BinaryStoreValueFactory.create(BinaryStoreValueFactory.java:244)
... 164 more

will be thrown in all other threads (with different SHA-1).

### DIFF
--- a/modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/S3BinaryStore.java
+++ b/modeshape-jcr/src/main/java/org/modeshape/jcr/value/binary/S3BinaryStore.java
@@ -221,7 +221,8 @@ public class S3BinaryStore extends AbstractBinaryStore {
     @Override
     public BinaryValue storeValue(InputStream stream, boolean markAsUnused) throws BinaryStoreException {
         // Cache file on the file system in order to have SHA-1 hash calculated
-        BinaryValue cachedFile = fileSystemCache.storeValue(stream, markAsUnused);
+        // File is marked as used (unused=false)
+        BinaryValue cachedFile = fileSystemCache.storeValue(stream, false);
         try {
             // Retrieve SHA-1 hash
             BinaryKey key = new BinaryKey(cachedFile.getKey().toString());


mbklein:
@coraxs After cherry-picking this change into modeshape 5.3.0 and building fcrepo-4.7.4 with it, I still get the "Unable to find binary value" error above.

coraxs:
Are you reproducing same case? Are you trying to  store same content from all threads (with same sha-1)? In that case flag unused is unimportant, because cleaning will occur anyway after first storing of content.

mbklein:
Sorry for the (very) late followup – I think my previous build process must have been flawed, because I actually did rebuild it and it's been working great. Thanks!

weebl2000:
Is there an issue for this PR on ModeShape project in JBoss JIRA? That way we can provide accurate changelog if a new version of modeshape is released without having to reference github.

mbenson:
@bbranan could you comment on this PR? Thanks!

bbranan:
@mbenson this looks good to me. The explanation that @coraxs provides to explain why this change is needed in a multi-threaded scenario makes sense. Given that there is a finally block attached to the try/catch immediately following this call which will set the file to unused and attempt to delete it suggests that setting the file to "used" in all cases is unlikely to result in cached files sticking around on the file system unintentionally.

mbenson:
note that @blafond opened MODE-2738 for this

awoods:
As a note, the Fedora project has released a fork of ModeShape-5.4.1.Final with the addition of this PR.
https://github.com/fcrepo4-exts/modeshape/releases/tag/modeshape-5.4.1.fcr

Once ModeShape releases, we will switch back over to the official ModeShape dependency.

PR #1672, opened 2018-08-01
MODE-2730 account for line ending differences to pass tests on Windows

mbenson:


mbenson:
Confirm that this still builds properly for me on Windows.

PR #1680, opened 2018-11-16
MODE-2720 Lucene index provider does not correctly handle LIKE constraints containing a backslash

mbenson:
Fix, replace #1667


blafond:
thx Matt local build works for me too.

PR #1684, opened 2018-12-08
fall back to default AWS credentials provider chain when accessKey/se…

mbenson:
…cretKey not specified


mbenson:
I was afraid this one might not merge cleanly. Addressing...

PR #1685, opened 2018-12-14
Add a version to the top level pom.xml.

cunningt:
I'm trying to mvn:versions:set on the project, but it doesn't look like I can if there is no version in the top level pom.xml.

[ERROR] Failed to execute goal org.codehaus.mojo:versions-maven-plugin:2.7:set (default-cli) on project modeshape: Project version is inherited from parent. -> [Help 1]

Adding a version cures that.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,6 +10,7 @@
 
 	<groupId>org.modeshape</groupId>
 	<artifactId>modeshape</artifactId>
+        <version>5.5-SNAPSHOT</version>
 
 	<packaging>pom</packaging>
 	<name>ModeShape</name>


mbenson:
did you try instead running `versions:set` using `-pl modeshape-parent`? With your change, does the version of the parent module get set properly? The project structure is definitely weird; I'll give you that.

mbenson:
After a bit more review, I think you're on the right track but that with not much more work we could get the project structure a little closer to sanity. I'll send another PR and link it.

mbenson:
see #1687

PR #1687, opened 2018-12-31
Maven refactor

mbenson:


cunningt:
@mbenson Thanks for doing this!   

Could we make one more addition here?    I found that it's also necessary to add a version to deploy/pom.xml if we want to versions:set on the project

https://github.com/ModeShape/modeshape/blob/master/deploy/pom.xml#L10-L12

Would it be possible to add a version tag there as well as part of this maven refactor?

mbenson:
@cunningt that last commit should have addressed that problem. Can you confirm?

cunningt:
@mbenson I think that fixes my issue.    Thanks!

